PR #15, opened 2020-06-05
Add Metafacture source dependency

blackwinter:
TBH, I'm not really happy with the newly introduced metafacture-core master-SNAPSHOT dependency (6382371). What about leveraging [Gradle source dependencies](https://blog.gradle.org/introducing-source-dependencies) instead?


fsteeg:
Nice, thanks! Didn't know about Gradle source dependencies, that's perfect.

Issue #5, opened 2019-09-02
Implement functions


fsteeg:
Change field values, use `metamorph.functions` ([wiki](https://github.com/metafacture/metafacture-core/wiki/Metamorph-functions)) as Fix [functions](https://github.com/LibreCat/Catmandu/wiki/Functions) like `replace_all(title,"My (.*) Pony","Our $1 Fish")`

fsteeg:
Implemented in https://github.com/metafacture/metafacture-fix/commit/6cdf1ebdc9dbdab66e3525b3dcfb68d9aa52d060 and merged via https://github.com/metafacture/metafacture-fix/pull/32. Closing.

PR #16, opened 2020-07-17
WIP: track changes for oersi-etl

fsteeg:
This is a WIP-PR to track the changes in metafacture-fix used in oersi-etl: https://gitlab.com/oersi/oersi-etl

Sample using these Fix features: https://gitlab.com/oersi/oersi-etl/-/blob/develop/data/to-oersibackend/oernds-to-oersi.flux

For merging these, it might make sense to cherry-pick into new, individual PRs for specific issues (like https://github.com/metafacture/metafacture-fix/issues/5).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,7 @@ subprojects {
   }
 
   group = 'org.metafacture'
-  version = '0.2.0-SNAPSHOT'
+  version = '0.2.0-rc12'
 
   apply plugin: 'checkstyle'
   apply plugin: 'eclipse'


blackwinter:
Can the first commit (683c247) be merged? Either before #60 is merged or right after. I think it would be valuable to have a stable identifier going forward.

fsteeg:
> Can the first commit ([683c247](https://github.com/metafacture/metafacture-fix/commit/683c247dfbeb213d4a6ecb9b4829071e6b422907)) be merged? Either before #60 is merged or right after. I think it would be valuable to have a stable identifier going forward.

Right, thanks, good catch, that was the plan, but I forgot. Added to 35-metafix in https://github.com/metafacture/metafacture-fix/pull/60/commits/1fe336bf81f8c490d20f2ffd7bfa53ad93063486.

blackwinter:
BTW (in the context of 683c247): Is there a reason why we don't just rename the subproject directories, mimicking the layout in metafacture-core?

- `org.metafacture.fix/` -> `metafix/` (no need to override the project name anymore)
- `org.metafacture.fix.ide/` -> `metafix-ide/`
- `org.metafacture.fix.vsc/` -> `metafix-vsc/`
- `org.metafacture.fix.web/` -> `metafix-web/`

fsteeg:
> BTW (in the context of [683c247](https://github.com/metafacture/metafacture-fix/commit/683c247dfbeb213d4a6ecb9b4829071e6b422907)): Is there a reason why we don't just rename the subproject directories, mimicking the layout in metafacture-core?

Sounds good. The current naming is the Xtext/Eclipse convention, but it should work to set up MWE workflow etc. for the simpler directories/names.

Issue #20, opened 2020-10-07
Test Fix language server usage from Visual Studio Code


fsteeg:
Like https://github.com/metafacture/metafacture-fix/issues/17, but for a Visual Studio Code extension, see e.g. https://www.typefox.io/blog/building-a-vs-code-extension-with-xtext-and-the-language-server-protocol

fsteeg:
Implemented by @katauber in https://github.com/metafacture/metafacture-fix/pull/23, see https://github.com/metafacture/metafacture-fix#visual-studio-code-extension. Usage with production Fix blocked by https://github.com/metafacture/metafacture-fix/issues/24, will do functional review with production Fix there. Closing.

Issue #17, opened 2020-10-07
Evaluate Xtext language server with Monaco widget


fsteeg:
To evaluate using an Xtext-generated language server for Fix (and later Flux) in a web playground (and later IDE/editor):

- [x] Set up an Xtext project that generates a language server +  Connect a Monaco code editor widget to the language server: (https://github.com/nmaoez/composer-language-server)
- [x] Use the existing Fix grammar in the language server project: [Grammar file](https://github.com/metafacture/metafacture-fix/blob/master/org.metafacture.fix/src/main/java/org/metafacture/fix/Fix.xtext)
- [x] Load [fix file](https://gitlab.com/oersi/oersi-etl/-/raw/develop/data/production/edu-sharing.fix) from OERSI ETL as example fix file -> changed to [fix file](https://github.com/metafacture/metafacture-fix#editor) because of problems with first one
- [ ] Deploy server and client on one of our servers for internal review

katauber:
Pushed current status to https://github.com/katauber/composer-language-server.

fsteeg:
Thank you @katauber. I ran it on my machine and saw it working. But I keep getting errors like `java.lang.IllegalStateException: This language server has already been initialized` after reload etc. Also in the unmodified master of the sample project. You also mentioned it does not work with multiple tabs connecting to the server. I took a short look at the server code and there is quite a lot of implementation of central network communication in that sample project. Following this route, we'd probably be working a lot on building a robust server setup.

I kind of expected something different going into this whole thing: a full working server implementation generated by Xtext, which the client just has to talk to. While there is a working server generated by Xtext, it is not an out-of-the-box solution for our use case. Maybe it is much simpler when building a VS Code extension which talks to the Xtext language server via standard IO or something (we will try that in #20). But the use case we need for a metafacture-playground (#19) – connecting a code widget with an Xtext language server – seems to be rather bleeding edge. The other thing we learned here is that syntax coloring so far is not even part of the LSP, but defined on the client side in the Monaco editor config.

So given these two things, maybe it makes sense to actually keep out the LSP part at first for #19. Instead, we optimize the Monaco client side config for Fix to get nice syntax coloring. When we integrate running the Fix, we also get error messages, which we can display in the UI (#3). At a later point, we could then connect the whole thing to a language server – providing even better language integration on top, instead of using it as a starting point. I think this would allow us to focus on building a useful playground first. What do you think, @katauber, @acka47?

acka47:
[...]

PR #23, opened 2020-11-04
Add an vs code extension for fix

katauber:
First result of [Issue 20](https://github.com/metafacture/metafacture-fix/issues/20).

- install Visual Studio Code
- install Node.js (including npm)
- checkout metfacture-fix project
- in metafacture-fix execute:
   - ./gradlew installDist
   - ./gradlew installServer
- in vscode-extension execute:
   - npm install
- open vscode-extension folder in Visaul Studio Code
- launch vscode extension by pressing F5 (opens new window of Visual Studio Code)
- open new file (file-ending .fix) or open existing fix-file


acka47:
Could you please push the changes in a "20-vscode-extension" branch in this repo? That is our general process and it makes it easier for me to check it out and test it.

fsteeg:
Great, that works (I installed vscodium, it all worked the same way).

I think this is a very good way to proceed. Some thoughts/issues for that:

- [ ] As discussed, some editorconfig fixes are required (`./gradlew check` should pass).
- [ ] Where the `vscode-extension` should go: as discussed, I had this idea that only the server side stuff should go here, but that's probably due to general misconceptions I had about the LSP and its usage, which I also mentioned in https://github.com/metafacture/metafacture-fix/issues/17#issuecomment-720383402. So maybe even in general, but at least in this stage of the Fix development, I think it makes sense to add the extension here. I'd suggest we rename `vscode-extension` to `org.metafacture.fix.vsc`.
- [ ] The (renamed) `org.metafacture.fix.vsc/xtext-server/` directory which we create with `./gradlew installServer` should not be checked in and added to `.gitignore`.
- [ ] The content of (renamed) `org.metafacture.fix.vsc/.gitignore` should be moved (and adapted for) the repo root `.gitignore`.
- [x] Opening the extension in vsc and launching another instance hosting the extension felt way more lightweight and straightforward than it sounded. This is great and I think we can stay with that approach for functional review and initial usage.
- [ ] We should add a section to the README with setup instructions like in https://github.com/metafacture/metafacture-fix/pull/23#issue-515298800.

fsteeg (approved):
Great, thanks @katauber!

PR #32, opened 2020-12-01
Current state of Fix implementation

fsteeg:
This is the current state of the Fix language implementation from the `oersi` branch (see https://github.com/metafacture/metafacture-core/pull/329/files), without anything oersi-specific or the new work on the Fix language server and extension by @katauber. Things are getting a little messy with the `oersi` branch being required for creating a Fix extension that actually works with the oersi workflows, so I think this is a good time to bring over the current state of the actual Fix implementation to master.

@blackwinter Would be great if you could take a look. It's a big diff, so this is probably not really a pleasure to review. I should have never let this branch diverge so much. It's all work in progress though, so it might make sense to improve this after merging, in additional issues, as things come up. Of course, if you have things that would make sense to address right here that's great too.

### DIFF
--- a/org.metafacture.fix/src/main/java/org/metafacture/fix/Fix.xtext
+++ b/org.metafacture.fix/src/main/java/org/metafacture/fix/Fix.xtext
@@ -36,7 +36,7 @@ Else:
 ;
 
 Do:
-  'do' name = ValidID '(' ( params += (QualifiedName|STRING) ( ',' params += (QualifiedName|STRING) )* )? ')' BEGIN
+  'do' name = ValidID '(' ( params += (QualifiedName|STRING) ( ',' params += (QualifiedName|STRING) )* ','? )? ( options = Options )? ')' BEGIN
     elements += Expression*
   END
   'end'
--- a/org.metafacture.fix/src/main/java/org/metafacture/fix/Fix.xtext
+++ b/org.metafacture.fix/src/main/java/org/metafacture/fix/Fix.xtext
@@ -43,7 +43,7 @@ Do:
 ;
 
 MethodCall:
-  name = ValidID '(' ( params += (QualifiedName|STRING) ( ',' params += (QualifiedName|STRING) )* )? ( ',' options = Options )? ')'
+  name = ValidID '(' ( params += (QualifiedName|STRING) ( ',' params += (QualifiedName|STRING) )* ','? )? ( options = Options )? ')'
 ;
 
 Options:
--- a/org.metafacture.fix/src/main/java/org/metafacture/fix/Fix.xtext
+++ b/org.metafacture.fix/src/main/java/org/metafacture/fix/Fix.xtext
@@ -43,7 +43,11 @@ Do:
 ;
 
 MethodCall:
-  name = ValidID '(' ( params += (QualifiedName|STRING) ( ',' params += (QualifiedName|STRING) )* )? ')'
+  name = ValidID '(' ( params += (QualifiedName|STRING) ( ',' params += (QualifiedName|STRING) )* )? ( ',' options = Options )? ')'
+;
+
+Options:
+  keys += (QualifiedName|STRING) ':' values += (QualifiedName|STRING) ( ',' keys += (QualifiedName|STRING) ':' values += (QualifiedName|STRING) )*
 ;
 
 terminal fragment ESCAPED_CHAR:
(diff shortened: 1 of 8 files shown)


blackwinter (approved):
After a cursory look, the following questions spring to mind:

1. There's quite a bit of duplication/overlap between `org.metafacture.metamorph.Metafix` here and `org.metafacture.metamorph.Metamorph` in metafacture-core. Fixes/enhancements in the latter had (and will have) to be ported to the former. Would it be possible, even desirable, to tie them closer together (maybe inherit from a common base class)?
2. Should we relax the Checkstyle config a bit? I'm not particularly fond of having to disable `MagicNumber`, `MultipleStringLiterals`, `ClassDataAbstractionCoupling` and `ClassFanOutComplexity` all over the place. Instead we might increase their limits or drop (some of) them altogether.
3. Why is there (explicit) flushing (`flushWith`)? I was under the impression from our initial discussion that Metafix would operate in the context of a complete record and hide the internal field stream from the user. What's the plan here? TBH, this is a deal breaker for us (well, me) if Metafix is not going to be the relief that we were hoping for in this regard. (Also, there are still tests marked `@Disabled("Fix choose flushing")` that I would have expected to pass after flushing had been introduced.)

fsteeg:
Thanks for looking and this and for your comments! I'll mix up the order a bit for replying:

> 3: Why is there (explicit) flushing (flushWith)? I was under the impression from our initial discussion that Metafix would operate in the context of a complete record and hide the internal field stream from the user. What's the plan here? [...]

Yes, hiding the stream and making the data available in a record mode is still the plan. The approach I'm taking for implementing Fix is to first stick close to the Morph implementation in order to make everything from Morph available in Fix first, taking that as a baseline for improvements.

> 1: There's quite a bit of duplication/overlap between org.metafacture.metamorph.Metafix here and org.metafacture.metamorph.Metamorph in metafacture-core. [...] Would it be possible, even desirable, to tie them closer together (maybe inherit from a common base class)?

Yes, there is quite some overlap right now. As outlined above, I'd expect that to reduce as we move away from the internal stream-based field handling in Fix though. Whatever overlap remains after that could go into a common base class or a delegate, if we think that makes sense at that point.

> 2: Should we relax the Checkstyle config a bit? I'm not particularly fond of having to disable [rules] all over the place. Instead we might increase their limits or drop (some of) them altogether.

I kind of like to have them as reminders of things to look at while doing the initial development. When things have settled down, I agree this is no good state, because why set up rules we don't stick to. But the particular rules here came from you, so feel free to tweak them.

blackwinter:
> Yes, hiding the stream and making the data available in a record mode is still the plan.

Excellent! Glad to hear that :)

> Yes, there is quite some overlap right now. As outlined above, I'd expect that to reduce as we move away from the internal stream-based field handling in Fix though.

I see, great.

> I kind of like to have them as reminders of things to look at while doing the initial development. When things have settled down, I agree this is no good state, because why set up rules we don't stick to.

I agree. Just assumed that this has already settled down ;) Do you mean the project as a whole?

> But the particular rules here came from you, so feel free to tweak them.

Adapted from metafacture-core, as suggested by you ;) And since you're the one doing the main development at the moment, I'll leave it up to you to tweak now or defer until later. Just wanted to express my general discomfort with some of those settings.

fsteeg:
> I agree. Just assumed that this has already settled down ;) Do you mean the project as a whole?

Hm, I was kind of thinking of the initial implementation of the Fix language, having hooked into all Metafacture parts. I'll try to look into the `ClassDataAbstractionCoupling` and `ClassFanOutComplexity` warnings soon. Maybe there isn't even that much more to do. I'll try to sort out the branches (starting with merging this) and issues, so we get a clearer picture of what's actually left and how to proceed.

> Adapted from metafacture-core, as suggested by you ;) [...] I'll leave it up to you to tweak now or defer until later

Right, thanks, now I remember how that came to be :). Then I think it makes sense to have them consistent with metafacture-core, will defer changing the settings for now, and look into fixing the issues. Later we can see if we want to tweak some settings in both repos.

blackwinter:
> Then I think it makes sense to have them consistent with metafacture-core

It's not active right now: metafacture/metafacture-core@8b98d06.

fsteeg:
> It's not active right now: metafacture/metafacture-core@8b98d06

Ah, right, thanks. So we can set up sensible settings here and then move them over to core.

PR #36, opened 2021-01-05
Extensions support of 'do'-bind

### LINKED ISSUE #27: Fix-Extension: Add auto-close functionality (do ... end)
When entering a function or a collector that starts with `do`, the editor should auto close with `end`. 
The `end` should emerge in the next line on the same level as the do.

e.g.;
```
do array('sourceOrganization')
 do entity('')
  add_field('type', 'Organization')
  do map('node.properties.ccm:university_DISPLAYNAME[].1', 'name')
    not_equals(string: '')
    not_equals(string: '- Alle -')
  end
 end
end
```

katauber:
New:
 - snippet for do-bind
 - workflow for auto deploy on metafacture.github.io

Will resolve #27


fsteeg:
This seems to contain a lot of unrelated changes, could you update it (e.g. by merging the master)?

fsteeg (approved):
Sorry, forgot that we are introducing the GitHub Action with this feature. Looks good to me. Feel free to squash some of the commits before moving to functional review and merging.

PR #44, opened 2021-01-14
Add workflow badge "Build and Deploy"

dr0i:
Complements Github Actions migration.

See https://github.com/hbz/lobid/issues/433.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ See [https://github.com/elag/FIG](https://github.com/elag/FIG)
 
 # State
 
-[![Build Status](https://travis-ci.org/metafacture/metafacture-fix.svg?branch=master)](https://travis-ci.org/metafacture/metafacture-fix)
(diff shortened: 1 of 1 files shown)


dr0i:
Not sure if this is the proper badge, because there are two workflows:
1. Build and Deploy
2. Build

While this PR covers 1 it obviously does not 2 - because this PR doesn't show up at  [1](https://github.com/metafacture/metafacture-fix/actions?query=workflow%3A%22Build+and+Deploy%22) but only at [2](https://github.com/metafacture/metafacture-fix/actions?query=workflow%3ABuild).

dr0i:
Same question as in https://github.com/metafacture/metafacture-flux/pull/11:
> Ah, I see: `Build` builds on every push, regardless of the name of the branch.
`Build and Deploy` does run only when pushing into the master (aka `main`) branch.
We may need then two badges, right?

dr0i:
On a second thought: it's a badge at the README and we want to reflect the status of the main/master branch there. No second badge is needed.

katauber (commented):
+1

katauber (approved):

dr0i:
Merged, closed.

PR #72, opened 2021-11-12
Introduce `Record` class.

### LINKED ISSUE #64: Create a Record class that wraps the `Map<String, Object>`
Came up in https://github.com/metafacture/metafacture-fix/pull/60#pullrequestreview-791798200.

blackwinter:
Resolves #64.


blackwinter:
Essential functionality should be ready for review now. There are still some missing pieces (esp. unit tests and Javadoc for newly introduced APIs), but I consider them optional for the purposes of this pull request (weren't present before either).

A slightly more severe issue might be that I went a little overboard with the initial assignment ;) And I did it in the wrong order, so it's not easily ripped out now. If you feel that the `Value` refactoring in particular should not have happened (either in the context of this pull request or at all), I'll start a new pull request from scratch.

Also, there are probably more opportunities for consolidation/simplification that I wasn't able to investigate yet.

If, however, there are no objections at all by any chance, feel free to merge right away in order to unblock further development.

blackwinter:
Sorry for the noise, I'll leave it alone now!

fsteeg (approved):
Great, thanks! The `Value` abstraction looks good. Just irritated a bit by the naming of `Array` and `Hash` for List and Map. Also `field` for key, I'd consider a field a key-value pair: field name and field value. However, maybe it's good to have more specific / domain / Catmandu-related naming at this level.

blackwinter:
Wow, that was fast :) Thanks!

> However, maybe it's good to have more specific / domain / Catmandu-related 
> naming at this level.

Exactly, that's what I was going for. But we can certainly discuss more.

PR #76, opened 2021-11-17
Remove Checkstyle `@SuppressWarnings`, disable for tests

### LINKED ISSUE #62: Agree on a single Checkstyle suppression mechanism
Came up in https://github.com/metafacture/metafacture-fix/pull/60#discussion_r737447591. I seem to remember that something was not possible with the original `checkstyle-disable-line` when adding support for `@SuppressWarnings` in https://github.com/metafacture/metafacture-fix/commit/d70d5d3b1405ab845c90f5ed9e4b76a52d73db02. Maybe I didn't see a way to suppress `MultipleStringLiterals` for the entire class? See https://github.com/metafacture/metafacture-fix/pull/60#discussion_r737447591 for other pro/cons of both approaches.

fsteeg:
Will resolve #62.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -5,7 +5,7 @@
 <module name="Checker">
     <module name="SuppressionSingleFilter">
         <property name="checks" value=".*"/>
-        <property name="files" value="xtext-gen"/>
+        <property name="files" value="xtext-gen|test"/>
     </module>
     <module name="SuppressWarningsFilter" />
     <module name="TreeWalker">
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -7,7 +7,6 @@
         <property name="checks" value=".*"/>
         <property name="files" value="xtext-gen|test"/>
     </module>
-    <module name="SuppressWarningsFilter" />
     <module name="TreeWalker">
         <module name="AbstractClassName"/>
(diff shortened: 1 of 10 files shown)


blackwinter:
Just had a quick look.

1. I assumed we would only disable specific checks for the tests.
2. The Checkstyle config for the annotation should be removed if it's no 
longer used.

fsteeg:
> 1. I assumed we would only disable specific checks for the tests. 

I did look into that again yesterday, but there seems to be no straightforward way to do that. One option would be to have separate configs for main and test, but that would duplicate a lot. As mentioned in https://github.com/metafacture/metafacture-fix/issues/62#issue-1045845269 and in our meeting, I think that was my reason to introduce annotation support, as it provides a convenient way to disable specific checks for a file / class. But I'm not picky here, it would all be fine with me: duplicate / separate configs, off for all tests (current state), or some proper setup in Checkstyle that I didn't find / get.

> 2. The Checkstyle config for the annotation should be removed if it's no longer used.

Right, thanks. After removing it, I also found missing annotations that didn't match my simple search pattern (multiple checks disabled). Pushed both in 632802a.

blackwinter:
You should be able to specify multiple `SuppressionSingleFilter`s with 
different `checks` for different `files`.

fsteeg:
Oh, right, thanks, how did I not find that? Pushed in ea789ff.

Issue #66, opened 2021-11-08
Make the full record available in binds


fsteeg:
See https://github.com/metafacture/metafacture-fix/pull/60#pullrequestreview-795591354.

Functional review: @TobiasNx
Code review: @blackwinter

blackwinter:
Copying previous discussion for better overview:

---

This prevents access to the actual record. Is this intentional?

_Originally posted by @blackwinter in https://github.com/metafacture/metafacture-fix/pull/60#discussion_r741215149_

---

My idea was that in the scope of the bind, the record (as the subexpressions see it) should only contain the bound value. I'm not sure why though, that might very well make no sense.

_Originally posted by @fsteeg in https://github.com/metafacture/metafacture-fix/pull/60#discussion_r743702989_

---

I wasn't able to find it documented anywhere, but Catmandu seems to preserve access to the whole record:

```perl
$fixes = <<EOF;
do list(path:foo,var:loop)
  copy_field(test,loop.baz)
end
EOF

$fixer = Catmandu::Fix->new(fixes => [$fixes]);

is_deeply $fixer->fix({foo => [{bar => 1}, {bar => 2}], test => 42}),
    {foo => [{bar => 1, baz => 42}, {bar => 2, baz => 42}], test => 42},
    'binding scope testing';
```

_Originally posted by @blackwinter in https://github.com/metafacture/metafacture-fix/pull/60#discussion_r743730670_

---

I agree, the full record should be available. I've opened https://github.com/metafacture/metafacture-fix/issues/66 since this does require some work that I probably won't have time to do this week (simply using the full record with the current implementation results in overwriting previous values under the same key, which is covered by a test).

_Originally posted by @fsteeg in https://github.com/metafacture/metafacture-fix/pull/60#discussion_r744725815_

blackwinter:
Catmandu's behaviour seems to [depend on the presence of the `var` option](https://github.com/hbz/Catmandu/blob/bb54964/lib/Catmandu/Fix/Bind/list.pm#L47):

```perl
$fixes = <<EOF;
do list(path:foo)
  copy_field(test,baz)
end
EOF

$fixer = Catmandu::Fix->new(fixes => [$fixes]);

is_deeply $fixer->fix({foo => [{bar => 1}, {bar => 2}], test => 42}),
    {foo => [{bar => 1}, {bar => 2}], test => 42},
    'binding scope w/o var testing';
```

See also hbz/Catmandu@22d7359.

[...]

PR #78, opened 2021-11-25
Handle and access repeated fields as arrays

### LINKED ISSUE #65: Consistent handling of repeated fields / arrays
Explicit arrays (e.g. in JSON input) vs. explicit / repeated values (e.g. in XML input).

Field names like:

`metadata.mods.identifier.value` vs.
`metadata.mods.identifier.*.value` vs.
`metadata.mods.identifier[].*.value`

See:

- https://github.com/metafacture/metafacture-fix/issues/35#issuecomment-932020224
- https://github.com/metafacture/metafacture-fix/pull/60#discussion_r738381260

fsteeg:
Will resolve #65.

This is far from consistent and a bit messy, but hopefully some steps in the right direction. I noted some TODOs which would result in some refactoring work (mainly around avoiding switch statements by moving functionality into enum elements; as well as consistent handling of special path segments like wildcards and indexes), but to make sure we're moving into the right direction, I'd like to focus on use cases right now.

For [our OERSI use case](https://gitlab.com/oersi/oersi-etl/-/merge_requests/99) (updated for this state in [8b27e98c](https://gitlab.com/oersi/oersi-etl/-/merge_requests/99/diffs?commit_id=8b27e98ceaeb1926ddb8359697269c4d03c61b16)), this provides some progress, but I don't think a functional review makes sense at this point. For that, I'd like to continue with the bind scope issues, which are related to https://github.com/metafacture/metafacture-fix/issues/66. To avoid a larger and longer-running branch with these additional changes, I think it makes sense to review and merge the current work in progress at this point.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Value.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Value.java
@@ -148,6 +148,10 @@ public class Value {
         return value == null || value.isNull();
     }
 
+    private static boolean isNumber(final String s) {
+        return s.matches("\\d+");
+    }
+
     public Array asArray() {
         if (isArray()) {
             return array;
@@ -292,6 +296,21 @@ public class Value {
             return list.toString();
         }
 
+        public void remove(final int index) {
+            list.remove(index);
+        }
+
+        private void removeNested(final String[] fields, final Value value) {
+            if (fields.length > 1 && isNumber(fields[1])) {
+                final int index = Integer.parseInt(fields[1]) - 1;
+                if (index >= 0 && index < value.asArray().size()) {
+                    if (value.asArray().get(index).isString()) {
+                        value.asArray().remove(index);
+                    }
+                }
+            }
+        }
+
     }
 
     /**
@@ -397,7 +416,18 @@ public class Value {
         private Value findNested(final String field, final String[] remainingFields) {
             final Value value = get(field);
 
-            // TODO: array of maps, like in insert nested
+            if (value.isArray()) {
+                if (remainingFields.length > 0 && isNumber(remainingFields[0])) {
+                    final int index = Integer.parseInt(remainingFields[0]) - 1;
+                    if (index >= 0 && index < value.asArray().size()) {
+                        final Value nestedValue = value.asArray().get(index);
+                        if (nestedValue.isString()) {
+                            return nestedValue;
+                        }
+                        // TODO: array of maps, like in insert nested
+                    }
+                }
+            }
 
             if (value.isHash()) {
                 return value.asHash().find(remainingFields);
@@ -467,7 +497,12 @@ public class Value {
                             }
                             break;
                         default:
-                            array.add(newHash(h -> h.insert(mode, remainingFields, newValue)));
+                            if (isNumber(remainingFields[0])) {
+                                array.add(new Value(newValue));
+                            }
+                            else {
+                                array.add(newHash(h -> h.insert(mode, remainingFields, newValue)));
+                            }
                             break;
                     }
                 }
@@ -499,7 +534,13 @@ public class Value {
                 remove(field);
             }
             else if (containsField(field)) {
-                get(field).asHash().removeNested(Arrays.copyOfRange(fields, 1, fields.length));
+                final Value value = get(field);
+                if (value.isArray()) {
+                    value.asArray().removeNested(fields, value);
+                }
+                if (value.isHash()) {
+                    value.asHash().removeNested(Arrays.copyOfRange(fields, 1, fields.length));
+                }
             }
         }
 
@@ -510,7 +551,7 @@ public class Value {
         }
 
         public void transformFields(final List<String> params, final UnaryOperator<String> operator) {
-            final String field = params.get(0);
+            final String field = params.get(0).replace(".*", "");
             final Value value = find(field);
 
             if (value != null) {
--- a/metafix/src/main/java/org/metafacture/metafix/Value.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Value.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
@@ -32,6 +33,8 @@ import java.util.stream.Stream;
  * or a {@link java.lang.String String}.
  */
 public class Value {
+
+    private static final String ASTERISK = "*";
 
     private final Array array;
     private final Hash hash;
@@ -203,14 +206,7 @@ public class Value {
     }
 
     public Value merge(final Value value) {
-        if (isHash() && value.isHash()) {
-            final Hash asHash = asHash();
-            value.asHash().forEach(asHash::put);
-            return this;
-        }
-        else {
-            return asList(a1 -> value.asList(a2 -> a2.forEach(a1::add)));
-        }
+        return asList(a1 -> value.asList(a2 -> a2.forEach(a1::add)));
     }
 
     @Override
@@ -237,6 +233,10 @@ public class Value {
         }
 
         return result;
+    }
+
+    static String[] tail(final String[] fields) {
+        return Arrays.copyOfRange(fields, 1, fields.length);
     }
 
     enum Type {
@@ -300,17 +300,61 @@ public class Value {
             list.remove(index);
         }
 
-        private void removeNested(final String[] fields, final Value value) {
-            if (fields.length > 1 && isNumber(fields[1])) {
-                final int index = Integer.parseInt(fields[1]) - 1;
-                if (index >= 0 && index < value.asArray().size()) {
-                    if (value.asArray().get(index).isString()) {
-                        value.asArray().remove(index);
-                    }
+        private void removeNested(final String[] fields) {
+            if (fields.length >= 1 && fields[0].equals(ASTERISK)) {
+                for (int i = 0; i < size(); ++i) {
+                    remove(i);
+                }
+            }
+            else if (fields.length >= 1 && isNumber(fields[0])) {
+                final int index = Integer.parseInt(fields[0]) - 1;
+                if (index >= 0 && index < size()) {
+                    remove(index);
                 }
             }
         }
 
+        public Value find(final String[] fields) {
+            Value result = null;
+            if (fields.length > 0) {
+                if (fields[0].equals(ASTERISK)) {
+                    result = find(tail(fields));
+                }
+                else if (isNumber(fields[0])) {
+                    final int index = Integer.parseInt(fields[0]) - 1;
+                    if (index >= 0 && index < size()) {
+                        final Value value = get(index);
+                        // TODO: move impl into enum elements, here call only value.find
+                        if (value != null) {
+                            switch (value.type) {
+                                case Hash:
+                                    result = value.asHash().find(tail(fields));
+                                    break;
+                                case Array:
+                                    result = find(tail(fields));
+                                    break;
+                                case String:
+                                    result = value;
+                                    break;
+                                default:
+                                    break;
+                            }
+                        }
+                    }
+                }
+                else {
+                    final Value newResult = newArray();
+                    forEach(c -> {
+                        newResult.asArray().add(c.asHash().find(fields[0])); /* TODO: non-hash */
+                    });
+                    result = newResult;
+                }
+            }
+            else {
+                result = new Value(this);
+            }
+            return result;
+        }
     }
 
     /**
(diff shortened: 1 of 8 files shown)


metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -499,25 +605,56 @@ private void removeNested(final String[] fields) {
...
-            findList(oldName, a -> a.forEach(v -> append(newName, v.toString())));
+            findList(oldName, a -> a.forEach(v -> appendValue(split(newName), v)));
+        }
+
+        private void appendValue(final String[] newName, final Value v) {
+            // TODO: impl and call just value.append
+            if (v != null) {
+                switch (v.type) {
+                    case String:
+                        append(Arrays.asList(newName).stream().collect(Collectors.joining(".")), v.asString());
```
blackwinter, line 638:
```suggestion
                        append(String.join(".", newName), v.asString());
```

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -438,47 +519,72 @@ public Value insert(final InsertMode mode, final String fieldPath, final String
...
-
-                    switch (remainingFields[0]) {
-                        case APPEND_FIELD:
-                            array.add(newHash(h -> h.insert(mode, nestedFields, newValue)));
+                if (value != null) {
+                    switch (value.type) {
+                        // TODO: move impl into enum elements, here call only value.insert
+                        case Hash:
+                            final String[] tail = tail(fields);
+                            final String[] rest = tail[0].startsWith("$") ? tail(tail) : tail; // TODO: why?
```
blackwinter, line 536: > TODO: why?

That's what I'd like to know as well ;)

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -499,25 +605,56 @@ private void removeNested(final String[] fields) {
...
+        }
+
+        private void appendValue(final String[] newName, final Value v) {
+            // TODO: impl and call just value.append
+            if (v != null) {
+                switch (v.type) {
+                    case String:
+                        append(Arrays.asList(newName).stream().collect(Collectors.joining(".")), v.asString());
+                        break;
+                    case Array:
```
blackwinter, line 640: Nothing to do?

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -499,25 +605,56 @@ private void removeNested(final String[] fields) {
...
         }
 
         public void transformFields(final List<String> params, final UnaryOperator<String> operator) {
             final String field = params.get(0);
             final Value value = find(field);
-
             if (value != null) {
-                removeNested(field);
-
+                removeNested(field.replace(".*", ""));
```
blackwinter, line 655: Why drop the wildcard(s)?

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -292,6 +296,65 @@ public String asString() {
...
+        }
+
+        private void removeNested(final String[] fields) {
+            if (fields.length >= 1 && fields[0].equals(ASTERISK)) {
+                for (int i = 0; i < size(); ++i) {
+                    remove(i);
+                }
+            }
+            else if (fields.length >= 1 && isNumber(fields[0])) {
+                final int index = Integer.parseInt(fields[0]) - 1;
```
blackwinter, line 310: Isn't Catmandu zero-based?

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -292,6 +296,65 @@ public String asString() {
...
+        }
+
+        public Value find(final String[] fields) {
+            Value result = null;
+            if (fields.length > 0) {
+                if (fields[0].equals(ASTERISK)) {
+                    result = find(tail(fields));
+                }
+                else if (isNumber(fields[0])) {
+                    final int index = Integer.parseInt(fields[0]) - 1;
```
blackwinter, line 324: Again, zero-based.

fsteeg:
> I would contend that it might be beneficial (crucial even) to collect (and discuss) an extensive number of tests illustrating the required use cases (a specification of sorts) - and then implement those, and only those.

I agree with that goal, and I was basically trying to go for that with the unit tests here. How can we better collect and discuss these use cases? Maybe something like really short fix snippets provided as use cases, also by @TobiasNx? Which we can run with Catmandu and Metafacture, and which we integrate as unit tests?

And thanks for the specific pointers here in the review, I'll work through them to clean this up some more. My main point for eventually merging this is that it fixes the current behavior of merging subsequent entities that have the same name.

blackwinter:
> I was basically trying to go for that with the unit tests here.

I'm sure you were. It's just that there seems to be quite a lot of implemented behaviour that's not actually covered by any tests. Some might be edge cases (nested find in array with wildcard: `a.*.c`, and without wildcard: `a.b.c`), others probably not so much (nested remove in array: `remove_field('name.*')`).

As long as we're unsure about the intended behaviour, we shouldn't implement those cases. Conversely, any implemented behaviour should manifest itself somehow in the tests; so we know for sure when and how it might get changed in the future.

blackwinter:
> My main point for eventually merging this is that it fixes the current behavior of merging subsequent entities that have the same name.

Would it be possible for you to reduce the pull request to the minimal requirements for your OERSI use case? Otherwise, I'll switch my vote to -0 as long as we're clear that any untested behaviour is considered experimental and subject to change (without notice).

blackwinter:
> How can we better collect and discuss these use cases? Maybe something like really short fix snippets provided as use cases, also by @TobiasNx? Which we can run with Catmandu and Metafacture, and which we integrate as unit tests?

Certainly, that would be great. Ideally even accompanied by equivalent Catmandu tests (cf. hbz/Catmandu@22d7359).

The need to discuss referred to behaviour that's not as clear-cut in (or not applicable to) Catmandu. As for the process, I don't have any ideas right now other than pull requests (here and in hbz/Catmandu).

blackwinter:
It might even make sense to port the Catmandu test suite in order to ensure compatibility. The next step would be a shared collection of (generic/implementation-independent) test cases which would be executed by both (all) implementations (maybe akin to [Parsing JSON is a Minefield: Testing Architecture](https://seriot.ch/projects/parsing_json.html#3)).

fsteeg:
Thanks again for your feedback, it helped me a lot, and I think this is starting to make sense. I've added TODOs / WDCDs (What Does Catmandu Do) for open questions.

> any untested behaviour is considered experimental and subject to change (without notice)

+1

To approach the WDCDs, maybe we should wrap up this PR and focus on starting a setup for testing as you outlined here:

> It might even make sense to port the Catmandu test suite in order to ensure compatibility. The next step would be a shared collection of (generic/implementation-independent) test cases which would be executed by both (all) implementations

I've always seen actual compatibility as a later step, but maybe it does make sense to start setting up something like that now. To get started, we could work on a modified version (with stuff like quoted variables as currently often required in Metafix etc.). It would still allow us to be guided by the Catmandu concepts, and provide a better way to work with examples (also for @TobiasNx, instead of working on full real-world examples as we currently do in OERSI).

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -302,71 +303,112 @@ public void remove(final int index) {
...
         private void removeNested(final String[] fields) {
             if (fields.length >= 1 && fields[0].equals(ASTERISK)) {
-                for (int i = 0; i < size(); ++i) {
-                    remove(i);
-                }
+                list.clear();
             }
             else if (fields.length >= 1 && isNumber(fields[0])) {
-                final int index = Integer.parseInt(fields[0]) - 1;
+                final int index = Integer.parseInt(fields[0]) - 1; // TODO: 0-based Catmandu vs. 1-based Metafacture
```
blackwinter, line 309: I can't think of anything off the top of my head. Where is Metafacture 1-based? Seeing as it doesn't have a concept of arrays...

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -695,26 +698,26 @@ public String asString() {
...
-                    final Value newValue = new Value(value);
-                    hash.put(field, oldValue == null ? newValue : oldValue.merge(newValue));
-                }
-            };
+    private enum InsertMode {
 
-            abstract void apply(Hash hash, String field, String value);
+        REPLACE {
+            @Override
+            void apply(final Hash h, final String field, final String value) {
```
blackwinter, line 707: Minor: Not particular fond of shortening the variable name.

blackwinter (approved):
OK, so here's my -0 for the sake of progress.

I'll intentionally leave any unresolved conversations (TODOs) open.

I still believe that #65 should not be closed as long as we're unclear about the big picture.

metafix/src/main/java/org/metafacture/metafix/Value.java
fsteeg: I was probably thinking of the way JsonDecoder numbers records and array elements.

blackwinter: Ah, okay, makes sense. Not sure, though, if both domains (names and indexes) _have_ to be congruent at all.

metafix/src/main/java/org/metafacture/metafix/Value.java
fsteeg: Avoiding the need to rename `hash` in https://github.com/metafacture/metafacture-fix/pull/78/commits/1904e0d5900f7410015a4a875a38cdb50a2b80db.

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -292,6 +296,65 @@ public String asString() {
...
+                            }
+                        }
+                    }
+                }
+                else {
+                    final Value newResult = newArray();
+                    forEach(c -> {
+                        newResult.asArray().add(c.asHash().find(fields[0])); /* TODO: non-hash */
+                    });
+                    result = newResult;
```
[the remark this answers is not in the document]
fsteeg: I think this is done, see: https://github.com/metafacture/metafacture-fix/blob/1904e0d5900f7410015a4a875a38cdb50a2b80db/metafix/src/main/java/org/metafacture/metafix/Value.java#L354.

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -438,47 +519,72 @@ public Value insert(final InsertMode mode, final String fieldPath, final String
         private Value insert(final InsertMode mode, final String[] fields, final String newValue) {
             final String field = fields[0];
 
-            if (fields.length == 1) {
+            if (fields.length == 1 && !fields[0].equals(ASTERISK)) {
                 mode.apply(this, field, newValue);
             }
             else {
                 if (!containsField(field)) {
                     put(field, newHash());
```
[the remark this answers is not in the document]
fsteeg: (Addressed first part: ~~Put ASTERISK?~~)

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -438,47 +519,72 @@ public Value insert(final InsertMode mode, final String fieldPath, final String
...
+                case LAST_FIELD:
+                    if (array.size() > 0) {
+                        final Value last = array.get(array.size() - 1);
+                        if (last.isHash()) {
+                            last.asHash().insert(mode, tail(fields), newValue);
+                        }
+                    }
+                    break;
+                default:
+                    if (isNumber(fields[0])) {
```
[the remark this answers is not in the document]
fsteeg: (Addressed `mode` part: ~~Also, why is the mode ignored here?~~)

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -391,19 +454,37 @@ private Value find(final String[] fields) {
...
+                        if (remainingFields[0].equals(ASTERISK)) {
+                            result = value.asArray().find(tail(remainingFields));
+                        }
+                        else {
+                            result =  value.asArray().find(remainingFields);
+                        }
+                        break;
+                    case Hash:
+                        if (remainingFields[0].equals(ASTERISK)) {
+                            result = value.asHash().find(tail(remainingFields));
```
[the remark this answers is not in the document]
fsteeg: (Addressed second part: ~~Let Hash.find() deal with wildcards.~~)

TobiasNx:
I do not know if the issue i found concerning creating arrays (with and without `set_array`) see #80 , is or should be part of this PR

metafix/src/main/java/org/metafacture/metafix/Value.java
fsteeg: (See https://github.com/metafacture/metafacture-fix/pull/78/files/5fa6e03b06324ab3259101a14d2038375bd2829a..4d956a7ac8c30dd9dafa998635d0b50271ccf9f8#r759169911)

metafix/src/main/java/org/metafacture/metafix/Value.java
fsteeg: (See https://github.com/metafacture/metafacture-fix/pull/78/files/5fa6e03b06324ab3259101a14d2038375bd2829a..4d956a7ac8c30dd9dafa998635d0b50271ccf9f8#r759169911)

fsteeg:
> I do not know if the issue i found concerning creating arrays (with and without set_array) see #80 , is or should be part of this PR

I think we should handle that separately. Merging this to avoid an even larger and longer-running branch. Expecting new pull requests after functional review of #65.

Issue #63, opened 2021-11-05
Emit internal `_` fields


fsteeg:
Came up in #60. Internal `_` fields should be emitted if explicitly listed in `retain`.

blackwinter:
Copying previous discussion for better overview:

---

"don't emit `_` fields" (24d72f0): Do you really think this is such a great idea? Why prevent anyone from mapping arbitrary fields? Who's to say there can't be a legitimate use case for this?

It should at least be possible to bypass this filter by somehow marking desired fields. E.g., all fields explicitly mapped by the user and/or listed in `retain()`. I guess it's okay to filter internal fields (such as `_id`), unless specifically requested for output.

_Originally posted by @blackwinter in https://github.com/metafacture/metafacture-fix/pull/60#discussion_r737620259_

fsteeg:
From the discussion: emit `_id` if specified in `retain()`.

Functional review: @TobiasNx
Code review: @fsteeg

TobiasNx:
Since I saw this in the catmandu documentation. It always seems to emit fields that have `_` and does not need additional specification in `retain()`:

See the example here `_id`:
https://librecat.org/Catmandu/#fixes

Also I think that `retain` should not be the function by choice since you need to specify all fields that you want to retain. All fields that are not stated are deleted. 

I think we should mimik the behaviour that keeps the field by default as in catmandu.

Test case:
I want to change something small in a record but keep the rest. Keeping fields with `_` by default would emit all fields. 
If I needed to expicitly state `"_id" ` in retain I would also need to state all the other fields at least the parent/root element of them that I need additionally. Otherwise I would loose all the other fields but `"_id"` oder that I forget.

If this is only about the metamorph internal fields (like `"_id"` are there even more?) and not all fields with `_` 
 I still would recommend a seperate function like `keep() ` additionally to `retain()` so that we do not have the deletion behaviour that is connected to `retain()`

blackwinter:
@TobiasNX: Ready for review. (GitHub won't let me assign ATM...)

blackwinter:
Oops, I may have been too quick with merging the pull request. Is the functional review still coming?

TobiasNx:
To be honest, I was planing to, but I realized  that I do not know how to review new MR for Fix, this seems different from metafacture when building from source.

blackwinter:
> I do not know how to review new MR for Fix

[...]

PR #83, opened 2021-12-01
Don't restrict bind scope to the bound var

### LINKED ISSUE #66: Make the full record available in binds

fsteeg:
Will resolve #66

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -17,7 +17,6 @@ jobs:
       run: ./gradlew build
     - name: Install language server
       run: ./gradlew installServer
-    - name: Change directory
-      run: cd metafix-vsc/
     - name: Install fix extension
+      working-directory: metafix-vsc/
       run: npm install
--- a/metafix/src/main/java/org/metafacture/metafix/RecordTransformer.java
+++ b/metafix/src/main/java/org/metafacture/metafix/RecordTransformer.java
@@ -92,11 +92,33 @@ class RecordTransformer {
     private void processBind(final Do theDo, final EList<String> params) {
         if (theDo.getName().equals("list")) { // TODO impl multiple binds via FixBind enum
             final Map<String, String> options = options(theDo.getOptions());
-            record.findList(options.get("path"), a -> a.forEach(value -> {
-                record.put(options.get("var"), value);
-                processSubexpressions(theDo.getElements());
-                record.remove(options.get("var"));
-            }));
+            record.findList(options.get("path"), a -> {
+                for (int i = 0; i < a.size(); ++i) {
+                    final Value value = a.get(i);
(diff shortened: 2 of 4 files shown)


blackwinter:
I'm in the process of introducing _virtual_ (internal) fields (for `_id`, see #63). Those could be leveraged here as well (add `var` as virtual field). I'll hold off with my PR until this one is merged.

blackwinter:
Catmandu only does this when the `var` option is given ([ref.](https://github.com/metafacture/metafacture-fix/issues/66#issuecomment-963272919)). OTOH, there's no test for a `list` bind without `var` right now.

Options:

1. Treat `list` bind without `var` as _undefined behaviour_ - subject to change (without notice).
2. Add support for both cases (including tests).
3. Disallow `list` bind without `var` for the time being (including tests).
4. Officially support current behaviour (including tests) - intentionally diverging from Catmandu.

My preference would be 2.

fsteeg:
> Catmandu only does this when the var option is given

I don't get that. What's the point in not making the record available, *in particular* if we bind nothing to a var? Could that be a bug in Catmandu?

In general, I'd like to focus on getting our use cases working, and not worry too much about Catmandu compatibility at this point, therefore I'd vote for 1:

> Treat list bind without var as undefined behaviour - subject to change (without notice)

blackwinter:
> What's the point in not making the record available,

That was the initial implementation, so there must have been _some_ point to it ;)

> in particular if we bind nothing to a var?

Then your scope is determined by the path, you just can't access the whole record. If you need that, bind to a variable.

> Could that be a bug in Catmandu?

As I said, I don't know if it's documented anywhere. But it was introduced as a feature enhancement (LibreCat/Catmandu@88dd6d1), the restricted scope was the original behaviour, AFAICS.

blackwinter (approved):

fsteeg:
Oh, that was fast. Here's what I wrote but waited to send after the build was fixed:

I've added access to the bound values to understand Catmandu's behavior in https://github.com/hbz/Catmandu/commit/093c6cb and added corresponding tests and implementation in https://github.com/metafacture/metafacture-fix/pull/83/commits/1b1ee8e.

I'll commit further tweaks based on your comments.

fsteeg:
Thanks for the quick review, I think your points should be addressed in https://github.com/metafacture/metafacture-fix/pull/83/commits/a70a7b886d97e5f6177a8cc3790ebd5c32664664.

I will merge once all conversations are resolved.

blackwinter:
Great, thanks! I think everything's resolved now.

PR #84, opened 2021-12-01
Emit all underscore fields.

### LINKED ISSUE #63: Emit internal `_` fields

blackwinter:
Introduced concept of _virtual_ (internal) fields which can be accessed like regular metadata fields, but aren't emitted by default.

Virtual fields are emitted if

1. explicitly retained with `retain("virtual_field")`, or
2. copied with `copy_field("virtual_field", "regular_field")`, or
3. added (i.e., overwritten) with `add_field("virtual_field", "value")`.

Turned `_id` (record ID) into a virtual field. Bind scope variables are obvious candidates as well (cf. #83).

Resolves #63.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Metafix.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Metafix.java
@@ -113,6 +113,7 @@ public class Metafix implements StreamPipe<StreamReceiver> {
     @Override
     public void startRecord(final String identifier) {
         currentRecord = new Record();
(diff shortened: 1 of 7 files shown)


blackwinter:
> Bind scope variables are obvious candidates as well (cf. #83).

Doesn't work without investing additional effort. Not sure if the concept of virtual fields is actually sustainable.

fsteeg (approved):

PR #91, opened 2021-12-09
Extend Vars support.

### LINKED ISSUE #88: Extend Vars support.
In order to allow all Fix expression parameters to include "compile-time" [variables](https://github.com/metafacture/metafacture-core/wiki/Metamorph-User-Guide#parameters-to-metamorph-definitions), I'd like to propose the following changes:

- Introduce new (custom) Fix functions (names up for discussion):
  - `put_var("varName", "value")`: Define single variable (with variable substitution).
  - `put_vars("varName": "value", ...)`: Define multiple variables (without variable substitution¹).
- Update existing expression usages:
  - `do`²/`if`/`elsif`/`unless`: Resolve variable substitutions in parameters.

¹ No support for variable substitution in options required yet.
² Currently there's no bind that accepts any parameters.

blackwinter:
Resolves #88.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Metafix.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Metafix.java
@@ -73,14 +73,25 @@ public class Metafix implements StreamPipe<StreamReceiver>, Maps { // checkstyle
     // TODO: Use SimpleRegexTrie / WildcardTrie for wildcard, alternation and character class support
     private Record currentRecord = new Record();
     private Fix fix;
-    private Map<String, String> vars = NO_VARS;
+    private Map<String, String> vars = new HashMap<>();
     private int entityCount;
     private StreamReceiver outputStreamReceiver;
     private String recordIdentifier;
     private List<Value.Hash> entities = new ArrayList<>();
 
     public Metafix() {
-        init();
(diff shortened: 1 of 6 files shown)


fsteeg (approved):

blackwinter:
That was quick :) Thanks!

metafix/src/main/java/org/metafacture/metafix/Metafix.java
```diff
@@ -247,6 +243,16 @@ public StreamReceiver getStreamReceiver() {
         return vars;
     }
 
+    public void putVar(final String key, final String value) {
+        try {
+            vars.put(key, value);
+        }
+        catch (final UnsupportedOperationException e) {
+            vars = new HashMap<>(vars);
```
blackwinter, line 251: @fsteeg: WDYT? The alternative would be to keep a separate map and `putAll(newVars)` in `Metafix(Reader, Map<String, String>)`, like Metamorph does (via `AbstractMetamorphDomWalker`).

blackwinter: Implemented above mentioned alternative in 887d6c6.

blackwinter:
Merging despite unresolved conversation. We can still revert/update later.

PR #97, opened 2021-12-20
Add support for Metamorph wildcards.

### LINKED ISSUE #89: Add support for Metamorph wildcards.
While Fix/Catmandu has its own variant of [wildcards](https://librecat.org/Catmandu/#wildcards) (which are only partially implemented currently), we also need [Metamorph wildcards](https://github.com/metafacture/metafacture-core/wiki/Metamorph-User-Guide#addressing-pieces-of-data) to be available (both are mutually compatible; see also [`WildcardTrie`](https://github.com/metafacture/metafacture-core/blob/1ecaa75a9c5c5f36ff40a76ad42872252576a297/metafacture-commons/src/main/java/org/metafacture/commons/tries/WildcardTrie.java)/[`SimpleRegexTrie`](https://github.com/metafacture/metafacture-core/blob/1ecaa75a9c5c5f36ff40a76ad42872252576a297/metafacture-commons/src/main/java/org/metafacture/commons/tries/SimpleRegexTrie.java)).

An alternative might be to implement [specialized fixes](https://metacpan.org/pod/Catmandu::MARC#MODULES) with [MARC paths](https://librecat.org/Catmandu/#marc-mab-pica-paths) instead (see also [MARCspec](https://marcspec.github.io/MARCspec/marc-spec.html)).

Functional review: @TobiasNx
Code review: @fsteeg

blackwinter:
Delegates `Value.Hash` retrieval to `SimpleRegexTrie` for pattern matching.

All operations involving field patterns must be translated to concrete field names.

NOTE: Requires (a lot) more tests. But existing tests, including previously disabled ones, are already providing some coverage - with _trivial_ (?) modifications to make them pass (not sure why they were testing the wrong order).

Resolves #89.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Metafix.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Metafix.java
@@ -71,7 +71,6 @@ public class Metafix implements StreamPipe<StreamReceiver>, Maps { // checkstyle
     private final Map<String, Map<String, String>> maps = new HashMap<>();
     private final StreamFlattener flattener = new StreamFlattener();
 
-    // TODO: Use SimpleRegexTrie / WildcardTrie for wildcard, alternation and character class support
     private Record currentRecord = new Record();
     private Fix fix;
     private Map<String, String> vars = new HashMap<>();
--- a/metafix/src/main/java/org/metafacture/metafix/Value.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Value.java
@@ -507,6 +507,7 @@ public class Value {
         private static final String FIELD_PATH_SEPARATOR = "\\.";
 
         private final Map<String, Value> map = new LinkedHashMap<>();
+        private final SimpleRegexTrie<String> trie = new SimpleRegexTrie<>();
 
         /**
          * Creates an empty instance of {@link Hash}.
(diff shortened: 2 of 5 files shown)


blackwinter:
The bulk of the effort went into `transformFields()`. There may still be some pitfalls left.

blackwinter:
Also, `transformFields()` and `find()`/`findNested()` are basically performing the same traversal operations. There may be room for consolidation.

fsteeg (approved):
Ah, nice, so we're using the trie for pattern matching only, not for storing the actual values. I was expecting this to require a larger change, but this looks very good :+1:

Is my understanding correct that right now, we're creating a new trie with a single value for every lookup? I think we could make the trie a member of `Hash` to avoid multiple instantiations.

I think the wrong order in the disabled tests was from a previous state, where we were not using `LinkedHashMap` for the record map.

blackwinter:
> Is my understanding correct that right now, we're creating a new trie with a single value for every lookup? I think we could make the trie a member of `Hash` to avoid multiple instantiations.

Yes, your understanding is correct - and yes, we might be able to make it a member. I had a (hypothetical) case where that would've failed, but that might've been before I started checking for the actual pattern (`trie.get(field).contains(pattern)`). I'll think about it a little more.

blackwinter:
Alright, I've changed the trie to a member (432ef88). Should be fine™ ;)

blackwinter:
> Ah, nice, so we're using the trie for pattern matching only, not for storing the actual values.

FTR, we can't use `SimpleRegexTrie` as a key/value store because it's meant for the reverse use case as a `Registry` for `NamedValueReceiver`s: It stores patterns (Metamorph lookups) and then retrieves values by concrete field name (metadata stream).

PR #101, opened 2021-12-23
Reduce test output.

blackwinter:
How invested are you in the current test output? I feel, as our test suite grows, it's becoming a little unwieldy. So I'd like to propose a more compact version.

Passing:

- Before (cropped): passing-before
- After: passing-after

Failing:

- Before (cropped): failing-before
- After: failing-after


fsteeg (approved):
Very nice! Shorter, *and* more useful output :+1:

Issue #100, opened 2021-12-22
Add more Fix functions.


blackwinter:
We're going to need some more Fix functions that aren't implemented yet. E.g.:

| Morph     | Fix           |
|-----------|---------------|
| `replace` | `replace_all` |
| `split`   | `split_field` |
| `unique`  | `uniq`        |

I'm just going to go ahead and implement those plus a bunch more from the [Catmandu Cheat Sheet](https://librecat.org/assets/catmandu_cheat_sheet.pdf).

Functional review: @TobiasNx
Code review: @fsteeg

TobiasNx:
In Catmandu `replace_all` does not seem to support regexp as in metafacture. That would be nice.

`unique` seem  to process only arrays in Catmandu. Metafacture is here also more powerful. 

`split_field` outputs arrays and differs from split-behavour in Metafacture.

Bonus: `Unique`/`uniq` objects not just strings would be awesome.

blackwinter:
> In Catmandu `replace_all` does not seem to support regexp as in metafacture. That would be nice.

What makes you think that? The cheat sheet says: `replace_all(title,'[au]','X')` -> `title: cXtmXndX`. And [the tests confirm](https://github.com/hbz/Catmandu/blob/dev/t/Catmandu-Fix-replace_all.t) this behaviour.

> `unique` seem to process only arrays in Catmandu. Metafacture is here also more powerful. [Bonus: `Unique`/`uniq` objects not just strings would be awesome.]

Well, yes, it seems that Metamorph's `unique` function is a little more powerful. But I don't think we've ever made use of that.

> `split_field` outputs arrays and differs from split-behavour in Metafacture.

Hm, other than the absence of arrays in Metamorph, I don't really see a difference. It just emits a repeated field instead of an actual array object.

But thanks for your input! Let's wait and see what our hybrid "better than both worlds" beast will look like in the end ;)

blackwinter:
@TobiasNx: Sorry, forgot to assign you for functional review.

TobiasNx:
This will take some time to review since this are a lot of functions

blackwinter:
That's understandable ;) Thanks for your diligence!

In effect, this means that #103 and #105 have passed code review, but won't be merged until both functional reviews are done - unless we decide otherwise.

fsteeg:
> In effect, this means that #103 and #105 have passed code review, but won't be merged until both functional reviews are done - unless we decide otherwise.

[...]

PR #110, opened 2022-01-19
Convert marked array entity into actual array object.

### LINKED ISSUE #109: Convert marked array entity into actual array object.
When receiving an entity from upstream that is marked as an array (e.g. from `JsonDecoder`), convert it into an actual `Value.Array` object instead of having to deal with different array representations internally.

Related to #65 and #100; maybe also #95, #96 and #106.

Functional review: @TobiasNx? or: n/a (refactoring)
Code review: @fsteeg

blackwinter:
The last commit (7d9b2a4) might be debatable; but I feel making intentions explicit works in our favor for now (the usual WDCD applies, though).

Note that I deliberately went in the direction of more explicit types instead of less:

> TODO? maybe this would be a good general internal representation, resulting
> in every value being either a hash or a string, no more separate array type.

Resolves #109 and fixes most of the array issues in #100 (see 11d1e71).

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Metafix.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Metafix.java
@@ -189,7 +189,9 @@ public class Metafix implements StreamPipe<StreamReceiver>, Maps { // checkstyle
             throw new IllegalArgumentException("Entity name must not be null.");
         }
 
-        final Value value = Value.newHash();
+        final Value value = name.endsWith(ARRAY_MARKER) ? Value.newArray() : Value.newHash();
+        // TODO: Remove array marker? => name.substring(0, name.length() - ARRAY_MARKER.length());
+
         addValue(name, value);
         entities.add(value);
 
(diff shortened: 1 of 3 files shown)


fsteeg (approved):
(Note to self & @TobiasNx: we'll probably have to add some `set_array` statements in OERSI & Playground samples when we deploy this.)

TobiasNx:
> (Note to self & @TobiasNx: we'll probably have to add some `set_array` statements in OERSI & Playground samples when we deploy this.)

@blackwinter & @fsteeg.
Does this mean new created fields cannot be part of an array unless the array already exists?

etc:
`add_field("ab.$append", "abc")
`move_field("xy", "qr.$append.cd")

Is this conceptually not possible or just not realised yet?

blackwinter:
> Does this mean new created fields cannot be part of an array unless the array already exists?

Exactly.

> Is this conceptually not possible or just not realised yet?

Intentionally not implemented (yet). Catmandu [supports it](https://github.com/hbz/Catmandu/blob/dev/t/Catmandu-Fix-add_field.t#L18-L20), though, so maybe we should too?

Issue #14, opened 2020-05-20
Facilitate extensibility (polyglot programming)


blackwinter:
* Allow functions to be defined in other languages than Fix (e.g. Java and/or JavaScript; cf. [Metamorph: Integration of Java and JavaScript](https://github.com/metafacture/metafacture-core/wiki/Integration-of-Java-and-JavaScript)).
* Might require an import mechanism in order to make those functions available (see #13).

blackwinter:
Can we make this "ready"? I'd like to work on it next, specifically the Java integration; JavaScript integration and import mechanism not needed (yet).

fsteeg:
> Can we make this "ready"? I'd like to work on it next, specifically the Java integration;

Would it make sense for @dr0i to do functional review here? I can do code review.

If @dr0i agrees, this should be ready then.

blackwinter:
> Would it make sense for @dr0i to do functional review here?

If that's possible, sure. Otherwise, I can do it with Limetrans (adapt [`hbz.limetrans.metamorph.ZDB`](https://github.com/hbz/limetrans/blob/master/src/main/java/hbz/limetrans/metamorph/ZDB.java) to be usable from Fix).

dr0i:
I could also give it a try and btw. learn about using fix. Just go on @blackwinter .

blackwinter:
Functional review: @dr0i
Code review: @fsteeg

blackwinter:
I'm always having trouble with the Xtext documentation :disappointed: @fsteeg: Do you happen to know where I can find the definition of `ValidID`?

fsteeg:
> Do you happen to know where I can find the definition of ValidID?

It seems to be defined as `'^'? ('a'..'z'|'A'..'Z'|'$'|'_') ('a'..'z'|'A'..'Z'|'$'|'_'|'0'..'9')*;` in `/org/eclipse/xtext/xbase/Xtype.xtext` (org.eclipse.xtext.xbase-2.17.0.jar).

blackwinter:
Thanks. So [use the source](https://github.com/eclipse/xtext-extras/blob/master/org.eclipse.xtext.xbase/src/org/eclipse/xtext/xbase/Xtype.xtext#L81), I guess...

It's just that I was under the assumption that qualified names were already allowed as function names, but I had to change `ValidID` to `QualifiedName`. So I wanted to know what `ValidID` actually is.

dr0i:
[...]

PR #124, opened 2022-02-01
Allow custom Java contexts to be called from Fix.

blackwinter:
Completes #14.


fsteeg (approved):
Nice extraction of the bind implementation, and a great way to do specific stuff in custom binds! :+1:

Issue #128, opened 2022-02-03
Start FIX Glossary


TobiasNx:
For consistent vocabulary we should set up a glossar:

Path-Wildcards: 
Array-Wildcards:

TobiasNx:
Resolved in #141

PR #139, opened 2022-02-09
Create glossary #128

TobiasNx:
Started the glossasry


TobiasNx:
Changes are copied to into #141

PR #140, opened 2022-02-09
Add infrastructure for integration tests.

blackwinter:
Allows for complete Flux pipelines to be run (and potentially also Catmandu in the future) - complementing our isolated unit tests. Hooks into our Gradle `check` task (and thus also into our GitHub checks). Can also be run directly, either with `./gradlew integrationTest` or with `metafix/integrationTest.sh`; both accept patterns as arguments to select specific tests to be run.

Inspired by https://github.com/TobiasNx/fix-FunctionalReview-Testing.

NOTE: It sure would have been nice to have this implemented as a native Gradle task instead of an external shell script, which then has to spawn a new Gradle process for each test. But unfortunately, I'm not as fluent in Groovy, so there ;) (Shouldn't be too hard, though. Maybe one day...)


fsteeg (approved):
Great, thanks for setting that up, looks very useful!

> NOTE: It sure would have been nice to have this implemented as a native Gradle task instead of an external shell script, which then has to spawn a new Gradle process for each test. But unfortunately, I'm not as fluent in Groovy, so there ;) (Shouldn't be too hard, though. Maybe one day...)

Wouldn't Java be an option too? I'd probably have set up a parameterized JUnit test class, providing the test case folders as parameters (or somehow generate them for all folders).

blackwinter:
> Wouldn't Java be an option too? I'd probably have set up a parameterized JUnit test class, providing the test case folders as parameters (or somehow generate them for all folders).

Huh, that never crossed my mind actually ;) Probably because - with the intent of elvolving @TobiasNx's [current setup](https://github.com/TobiasNx/fix-FunctionalReview-Testing/blob/c9d9447983fc435d01b35ac57046dc1dd3ddbb11/global-test.sh) - I framed it more as "execute some programs" (`flux`, `catmandu`, `diff`, ...) and "print some pretty output". Even though I've actually built [such a thing](https://github.com/hbz/limetrans/blob/master/src/main/java/hbz/limetrans/test/TransformationTestSuite.java) before :)

fsteeg:
> I framed it more as "execute some programs" (flux, catmandu, diff, ...)

Right, and that probably makes a lot of sense for that goal of a common matafix/catmandu test suite.

TobiasNx:
First try. I had to search a little bit for the integration test, but then I found them. In  my opinion they are quite convenient to set up as long as you are working in one type of transformation (e.g. json->json). Since the flux and the naming is consistent.

A first problem I spottet, I have to output everything in one line, see my commit all vacuum tests fail due to the "pretty" formating of mine.

blackwinter:
> First try. I had to search a little bit for the integration test, but then I found them.

Well, you could have looked at the code :) But by the same token, I _should_ have mentioned the location.

> A first problem I spotted, I have to output everything in one line, see my commit all vacuum tests fail due to the "pretty" formating of mine.

I don't understand, which commit? And why shouldn't you be able to output pretty results? The expected output just has to match.

blackwinter:
The pretty JSON for this example is

```json
{
  "key" : [ "value1", "value3" ]
}
```

not

```json
{
    "key": [
        "value1",
        "value3"
    ]
}
```

Nothing we can do about it; you have to prepare the expected output according to what would actually get produced by Metafacture.

Issue #119, opened 2022-01-31
`do list` does not work properly with path wildcards


TobiasNx:
[See here](https://metafacture.org/playground/?flux=PG_DATA%0A%7Cdecode-xml%0A%7Chandle-generic-xml%0A%7Cfix%0A%7Cencode-json%28prettyPrinting%3D%22true%22%29%0A%7Cprint%0A%3B%0A&fix=set_array%28%22coll%5B%5D%22%29%0Ado+list%28path%3A%22feld%3F%22%2C+%22var%22%3A%22%24i%22%29%0A++++add_field%28%22coll%5B%5D.%24append.feldtest%22%2C%22true%22%29%0A++++copy_field%28%22%24i.a.value%22%2C+%22coll%5B%5D.%24last.a%22%29%0Aend%0Aretain%28%22coll%5B%5D%22%29&data=%3C%3Fxml+version%3D%221.0%22+encoding%3D%22UTF-8%22%3F%3E%0A%3Crecord%3E%0A++++%3CfeldA%3E%3Ca%3EDog%3C/a%3E%3C/feldA%3E%0A++++%3CfeldA%3E%3Ca%3EApe%3C/a%3E%3C/feldA%3E%0A++++%3CfeldA%3E%3Ca%3EGiraffe%3C/a%3E%3C/feldA%3E%0A++++%3CfeldB%3E%3Ca%3ECrocodile%3C/a%3E%3C/feldB%3E%0A++++%3CfeldB%3E%3Ca%3EHuman%3C/a%3E%3C/feldB%3E%0A++++%3CfeldB%3E%3Ca%3EBird%3C/a%3E%3C/feldB%3E%0A%3C/record%3E&active-editor=fix)

Should be:
```
{
  "coll" : [ {
    "feldtest" : "true",
    "a" : "Dog"
  },{
    "feldtest" : "true",
    "a" : "Ape"
  },{
    "feldtest" : "true",
    "a" : "Giraffe"
  }, {
    "feldtest" : "true",
    "a" : "Crocodile"
  },
 {
    "feldtest" : "true",
    "a" : "Human"
  },
 {
    "feldtest" : "true",
    "a" : "Bird"
  }
 ]
}
```

TobiasNx:
do list-Test: https://github.com/TobiasNx/fix-FunctionalReview-Testing/commit/67375bfae47e7aab2a4eda4b9ef2e5a029acbdfb

blackwinter:
[...]

PR #142, opened 2022-02-10
Fix path wildcards in combination with arrays.

### LINKED ISSUE #115: Conditional `any_equal` not working correctly with path wildcards.
When using path wildcards (#89) to select multiple subfields, `any_equal` fails to match any of them (see 8777ffd for unit tests, and [Limetrans](https://github.com/hbz/limetrans/commit/b595fbbba5fbdb3a0cccaaa0cd4fb496f3ea6db2#diff-39fb43e4ba2206cf8b1e62f5dfd8457c8714d288ad298609e752e09246b777edR211-R228) for a real-world use case).

blackwinter:
Resolves #115 and #119.

NOTE: The updated test `MetafixRecordTest.appendWithAsteriksWildcardAtTheEnd()` was probably wrong in the first place; the original [comment](https://github.com/metafacture/metafacture-fix/issues/89#issuecomment-999433570) expected the other values as well, but the actual [review](https://github.com/TobiasNx/fix-FunctionalReview-Testing/blob/master/data/approved/wildCardTest/expected_test6.json#L13) (which was the basis for the unit test) did not.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Value.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Value.java
@@ -533,8 +533,10 @@ public class Value {
          */
         public Value get(final String field) {
             // TODO: special treatment (only) for exact matches?
-            final List<Value> list = matchFields(field, Stream::filter).map(this::getField).collect(Collectors.toList());
-            return list.isEmpty() ? null : list.size() == 1 ? list.get(0) : new Value(list);
+            final List<Value> list = findFields(field).map(this::getField).collect(Collectors.toList());
+            return list.isEmpty() ? null : list.size() == 1 ? list.get(0) : newArray(a -> list.forEach(v -> v.matchType()
+                        .ifArray(b -> b.forEach(a::add))
+                        .orElse(a::add)));
         }
 
         public Value getField(final String field) {
@@ -614,7 +616,7 @@ public class Value {
          * @param fields the field names
          */
         public void retainFields(final Collection<String> fields) {
-            map.keySet().retainAll(fields.stream().flatMap(f -> matchFields(f, Stream::filter)).collect(Collectors.toSet()));
+            map.keySet().retainAll(fields.stream().flatMap(this::findFields).collect(Collectors.toSet()));
         }
 
         /**
@@ -658,7 +660,11 @@ public class Value {
         }
 
         /*package-private*/ void modifyFields(final String pattern, final Consumer<String> consumer) {
-            matchFields(pattern, Stream::filter).collect(Collectors.toSet()).forEach(consumer);
+            findFields(pattern).collect(Collectors.toSet()).forEach(consumer);
+        }
+
+        private Stream<String> findFields(final String pattern) {
+            return matchFields(pattern, Stream::filter);
         }
 
         private <T> T matchFields(final String pattern, final BiFunction<Stream<String>, Predicate<String>, T> function) {
--- a/metafix/src/test/java/org/metafacture/metafix/MetafixBindTest.java
+++ b/metafix/src/test/java/org/metafacture/metafix/MetafixBindTest.java
@@ -598,25 +598,25 @@ public class MetafixBindTest {
     }
 
     @Test
-    @Disabled("See https://github.com/metafacture/metafacture-fix/issues/119")
+    // See https://github.com/metafacture/metafacture-fix/issues/119
     public void shouldIterateOverListWithCharacterClass() {
         shouldIterateOverList("n[ao]me", 3);
     }
 
     @Test
-    @Disabled("See https://github.com/metafacture/metafacture-fix/issues/119")
+    // See https://github.com/metafacture/metafacture-fix/issues/119
     public void shouldIterateOverListWithAlternation() {
         shouldIterateOverList("name|nome", 3);
     }
 
     @Test
-    @Disabled("See https://github.com/metafacture/metafacture-fix/issues/119")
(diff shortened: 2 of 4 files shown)


fsteeg (approved):
All these enabled test cases :raised_hands:

TobiasNx:
> Resolves #115 and #119.
> 
> NOTE: The updated test `MetafixRecordTest.appendWithAsteriksWildcardAtTheEnd()` was probably wrong in the first place; the original [comment](https://github.com/metafacture/metafacture-fix/issues/89#issuecomment-999433570) expected the other values as well, but the actual [review](https://github.com/TobiasNx/fix-FunctionalReview-Testing/blob/master/data/approved/wildCardTest/expected_test6.json#L13) (which was the basis for the unit test) did not.

I have checked that test, it is not wrong the new output seems wrong imo.

The comment suggested/asked how to handle asterisk at the end of an wild-card if it does include all variations of the whole path incl. the subfields. 

Now the behaviour seems weird. 

input is this:

```
{
	"animals": ["dog", "cat", "zebra"],
	"animal": "bunny",
	"animols": {
		"name": "bird",
		"type": "TEST"
	},
	"ANIMALS": "dragon and unicorn"

}
```

Fix:
```

copy_field("ani*", "stringimals[].$append")

set_array("strongimals[]")
copy_field("ani*", "strongimals[].$append")
```


ani* => as expecded would only match with the field `"animal"`

[my comment asked](https://github.com/metafacture/metafacture-fix/issues/89#issuecomment-999433570) if the behaviour of fix would be the same as morph. then it would match with "animals", "animal", "animols" and include other values.

[Now the result](https://github.com/TobiasNx/fix-FunctionalReview-Testing/blob/5a064c36df1d4342446f0a01c2210089246abe44/data/approved/wildCardTest/result_test6.json#L9) includes "animal" and "animals" and transforms them to simple string values and does some strange behaviour from the other tests since it does not provide index numbers in `"stringimals"` and adds `$append`.

If `"animals"`, `"animal"` and `"animols"` should be included, so not just simple `string`-fields but also objects and arrays, the expected would be:

```
  "stringimals" : {
    "1" : ["dog", "cat", "zebra"],
    "2" :  "bunny",
    "3" : {"name": "bird", "type": "TEST"}
  } ,
  "strongimals" : [ ["dog", "cat", "zebra"], "bunny" ,  {"name": "bird", "type": "TEST"}]
```

blackwinter:
> my comment asked if the behaviour of fix would be the same as morph. then it would match with "animals", "animal", "animols" and include other values.

No, those are two different things:

1. Do Fix path wildcards match the behaviour of Morph wildcards in that they extend to the _whole_ path?
2. What happens if a Fix path wildcard matches an array or a hash?

As for 1., we (well, I) decided against it (see [my answer](https://github.com/metafacture/metafacture-fix/issues/89#issuecomment-999444877) to your comment); Fix path wildcards only ever match a single path _segment_.

The behaviour for 2. is what changed here. Previously, the array (`animals[]`) would simply get stuck in a [dead end](https://github.com/metafacture/metafacture-fix/blob/ffa65cb5a1364207e2599017e70fa3d50460aab4/metafix/src/main/java/org/metafacture/metafix/Value.java#L594-L596), that's why IMO the test was wrong.

The pattern `ani*` matches all field names `animals[]`, `animal` and `animols`, so it's basically equivalent to

```perl
copy_field("animals[]", "stringimals[].$append")
copy_field("animal", "stringimals[].$append")
copy_field("animols", "stringimals[].$append")
```

The question then becomes whether each of these cases is implemented correctly (or at all). And `copy_field("array", "new_array[].$append")` already works, that's not new. Only the last case (`copy_field("hash", "new_array[].$append)`) [does not work](https://github.com/metafacture/metafacture-fix/blob/ffa65cb5a1364207e2599017e70fa3d50460aab4/metafix/src/main/java/org/metafacture/metafix/Value.java#L602) at the moment.

TobiasNx:
> > my comment asked if the behaviour of fix would be the same as morph. then it would match with "animals", "animal", "animols" and include other values.
> 
> No, those are two different things:
> 
>     1. Do Fix path wildcards match the behaviour of Morph wildcards in that they extend to the _whole_ path?
> 
>     2. What happens if a Fix path wildcard matches an array or a hash?

1. Yes that was the reason I modeled the test like this and not include subfields incl. array-fields.
2. Agreed. Also agree that the object/hash handling is missing.

But also how array and hash are handled is relevant. But this could be a problem of the fix and not of your changes.
(It is: https://metafacture.org/playground/?flux=PG_DATA%0A%7Cas-records%0A%7Cdecode-json%0A%7Cfix%0A%7Cencode-json%28prettyPrinting%3D%22true%22%29%0A%7Cprint%0A%3B&fix=copy_field%28%22a%5B%5D%22%2C+%22z%5B%5D.%24append%22%29&data=%7B%0A++++%22a%22%3A+%5B%22b%22%2C+%22c%22%2C+%22d%22%5D%2C%0A++++%22z%22%3A+%5B%5D%0A%7D&active-editor=fix)

`copy_field("animals[]", "stringimals[].$append")`

operates like I would expect `copy_field("animals[].*", "stringimals[].$append")`

blackwinter:
> But this could be a problem of the fix and not of your changes.

Exactly, this is independent of the changes discussed here as it's already working this way.

TobiasNx:
> > But this could be a problem of the fix and not of your changes.
> 
> Exactly, this is independent of the changes discussed here as it's already working this way.

my questions are solved. +1

PR #146, opened 2022-02-11
Include location and text in Fix execution exception.

### LINKED ISSUE #123: Error-Response - ideas
For users error-responses like that:

`org.metafacture.framework.MetafactureException: java.lang.IllegalStateException: expected String, got Hash`

do not help understanding where the error is in the fix.

It should state the function, ideally also the line in the fix.

blackwinter:
Resolves #123.


fsteeg (approved):
Nice, with line numbers and text!

PR #158, opened 2022-02-17
Introduce `@MetafixToDo` annotation to mark tests as disabled.

blackwinter:
Since we're keeping quite a few disabled tests in stock that indicate pending work (which is a good thing), it's not always clear (to me, at least) when any of those tests might be resolved due to (seemingly unrelated) fixes. So I'd like to propose a new `@MetafixToDo` annotation instead of the current `@Disabled` annotation, which takes care of reminding us of no longer failing tests: Failing tests marked as "to do" will be reported as *skipped* (just as before¹); passing tests marked as "to do" will be reported as *failed* (so we know we can remove the annotation).

¹ Except not quite: One problem with the proposed approach is that the log messages of still failing tests somewhat clutter the test output. We could reduce the log level for tests to `ERROR` as a remedy. OTOH, it's not entirely useless to see those messages, because sometimes disabled tests fail for the wrong reason ;) (see `MetafixSelectorTest.reject()` for instance).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ subprojects {
       'equalsverifier': '3.8.2',
       'jetty':          '9.4.14.v20181114',
       'jquery':         '3.3.1-1',
-      'junit_jupiter':  '5.4.2',
+      'junit_jupiter':  '5.8.2',
       'junit_platform': '1.4.2',
       'metafacture':    '5.3.1',
       'mockito':        '2.27.0',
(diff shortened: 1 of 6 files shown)


fsteeg (approved):
Nice! (Just today I was trying to find some way to run only disabled tests to check if a change makes any of them pass.)

blackwinter:
Haha, what a coincidence :)

blackwinter:
Since it's getting a bit unwieldy in day-to-day work after all, I've added (f4a7523) a system property `org.metafacture.metafix.disableToDo=<boolean>` to optionally disable all `@MetafixToDo`s, so they're equivalent to `@Disabled` before. Personally, I'm running `test` with this setting enabled and `check` with this setting disabled (unset), which appears to hit the sweet spot for me.

PR #133, opened 2022-02-04
Extract FixPath logic from Value

fsteeg:
As discussed in our meeting, this is an interim state for #102.

Issue #149, opened 2022-02-14
Add `delete:`-Attribute to lookup and change the default filter-modus to an optional configuration


TobiasNx:
In contrast to Catmandu lookup in metafacture fix is by default a filter.

Catmandu in contrast has the attribute `delete:`:

`lookup(color2,”lookup.csv”,delete:1)`

blackwinter:
See [Catmandu::Fix::lookup test](https://github.com/hbz/Catmandu/blob/dev/t/Catmandu-Fix-lookup.t#L18-L22).

TobiasNx:
@fsteeg can this be closed?

blackwinter:
AFAICS, should have been closed by #170 (e9dd4ab).

blackwinter:
Pending integration test:

- [lookupInSubfieldInArrayOfObjects_complex_2_withNotMatchingValue](https://github.com/metafacture/metafacture-fix/tree/master/metafix/src/test/resources/org/metafacture/metafix/integration/lookup/fromJson/toJson/lookupInSubfieldInArrayOfObjects_complex_2_withNotMatchingValue)

PR #170, opened 2022-02-28
Consolidate path traversal

### LINKED ISSUE #102: Consolidate path traversal.
Also, `transformFields()` and `find()`/`findNested()` are basically performing the same traversal operations. There may be room for consolidation.

_Originally posted by @blackwinter in https://github.com/metafacture/metafacture-fix/issues/97#issuecomment-998095139_

For now, I'm getting a bit uneasy about the divergence of object traversal between `find()`/`insert()` and `transformFields()`. I should probably work on that next.

_Originally posted by @blackwinter in https://github.com/metafacture/metafacture-fix/pull/98#pullrequestreview-838593348_

Right, this is also where I'm currently at in #95, where certain things only work in a bind if we first do e.g. an `upcase` (which calls `transformFields`).

_Originally posted by @fsteeg in https://github.com/metafacture/metafacture-fix/issues/98#issuecomment-1000266444_

fsteeg:
This is my suggestion to fix #102.

Based on the already merged extraction of FixPath (#133), here are some high-level changes:

- Make FixMethod interact with record, which interacts with FixPath (instead of creating FixPaths everywhere)
- Consolidate inserting values: replace append/replace methods with `insertInto` usage
- Implement transform methods based on `findIn` and `insertInto` to resolve the main issue of #102
- For cases where the `findIn` path does not work with `insertInto` (like `title-?` or `some.*.nested.*`), use a new `Value#path` field containing the full path to each value (created by `Metafix#flattener`)

This fixes some things that didn't work before (and changes the behavior for some things that did not work before and still don't work). See tests diff and issues mentioned in the commit messages.

I think this is an improvement over the current master state, and a good foundation for moving forward. Depending on the current, concrete requirements from our projects, this could be:

- Further improve behavior (e.g. handling of repeated fields, wildcards) by revisting the [issues that mention #102](https://github.com/metafacture/metafacture-fix/issues?q=is%3Aissue+is%3Aopen+%23102)
- Look into performance improvements (improvements in findIn/insertInto will benefit all transformations)
- Refactoring this to get it closer to the Catmandu terminology (e.g. `new FixPath(field, hash).get()`)


blackwinter (changes requested):
I'm sorry but there's just too much going on here. c25ede3 breaks our Limetrans workflow in a bunch of different ways (see below) and I'm officially giving up. Also, there seem to be only a few To Dos resolved, but more added. Not really the outcome we were hoping for, is it? (Integration test results as of 6d74bca: 3 fixed, 2 broken.)

So here are just some preliminary remarks and questions, this is not the final assessment.

----

Behaviour changes observed in Limetrans Alma mapping:

1. It's unclear where they're coming from but we get `null` values in records now: `Error while executing Fix expression [...]: vacuum()`, caused by `java.lang.IllegalStateException: Expected Array or Hash or String, got null`.
2. `uniq()` does not work anymore due to the inclusion of `path` in `Value.equals()`.
3. `lookup()` behaviour seems to have changed unexpectedly (see also #149).
4. Entity boundaries are violated (probably due to some error/change w.r.t. `$append`).

fsteeg:
Thank you for your review and for helping me make sense of all of this @blackwinter!

I've pushed my current state, which addresses many of the issues you brought up.

I'll continue with the remaining ones (`capitalize` & `append`/`prepend`).

blackwinter (commented):
Okay, thanks for the update.

fsteeg:
Changed the prepend/append/capitalize/trim tests to the original expected behavior and tweaked the implementation accordingly in be764e4.

Thanks for the additional feedback, addressed it as far as I could for now in 8b27a52.

blackwinter (changes requested):

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
```diff
@@ -424,10 +438,24 @@ public void shouldIgnoreOptionsOnSubsequentLookupInExternalFileMap() {
     }
 
     @Test
-    @MetafixToDo("Lookup with unknown internal map should not touch record")
     public void shouldNotLookupInUnknownInternalMap() {
-        assertMap(
-                LOOKUP + " 'testMap')"
+        MetafixTestHelpers.assertFix(streamReceiver, Arrays.asList(
+                LOOKUP + " '" + CSV_MAP + "', sep_char:'\t')"
```
blackwinter, line 443: This is not an _internal_ map.

metafix/src/main/java/org/metafacture/metafix/FixPath.java
```diff
@@ -60,6 +60,7 @@ private FixPath(final String[] path) {
     }
 
     /*package-private*/ Value findIn(final Array array) {
+
```
blackwinter, line 63: I didn't suggest that.

blackwinter (changes requested):

metafix/src/main/java/org/metafacture/metafix/FixPath.java
fsteeg: I thought that was implied by:

> Empty lines at start and end of a class body should be consistent; either both (usually) or none (for short bodies).

And:

> (another example would be FixPath.findIn() in this pull request)

Anyway, please add blank lines as you like, I have no particular preference.

blackwinter: > I thought that was implied by:
> 
> > Empty lines at start and end of a class body should be consistent; either both (usually) or none (for short bodies).

This referred to _class bodies_.

> And:
> 
> > (another example would be FixPath.findIn() in this pull request)

We were talking about empty lines in general at this point. What I meant was:

```diff
     /*package-private*/ Value findIn(final Hash hash) {
         final String currentSegment = path[0];
         final FixPath remainingPath = new FixPath(tail(path));
-
         if (currentSegment.equals(ASTERISK)) {
             // TODO: search in all elements of value.asHash()?
             return remainingPath.findIn(hash);
         }
-
         final Value value = hash.get(currentSegment);
         return value == null || path.length == 1 ? value : value.extractType((m, c) -> m
                 .ifArray(a -> c.accept(remainingPath.findIn(a)))
```

> Anyway, please add blank lines as you like, I have no particular preference.

Okay, will do (as I come across them).

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
fsteeg: Right, thanks, fixed in [6f455eb](https://github.com/metafacture/metafacture-fix/pull/170/commits/6f455ebedd2d848537036e7e05075accf14a4b64).

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
```diff
@@ -33,6 +33,7 @@
  * @author Fabian Steeg
  */
 @ExtendWith(MockitoExtension.class)
+@ExtendWith(MetafixToDo.Extension.class)
```
blackwinter, line 36: This is no longer needed now.

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -313,7 +313,8 @@ public void orElseThrow() {
 
     private abstract static class AbstractValueType {
 
-        protected static final Predicate<Value> REMOVE_EMPTY_VALUES = v -> v.extractType((m, c) -> m
+        protected static final Predicate<Value> REMOVE_EMPTY_VALUES = v ->
+            v.isNull() ? true : v.extractType((m, c) -> m
```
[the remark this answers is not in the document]
fsteeg: The null handled here is actually coming from the [explicit null setting for non-found values to be deleted after lookup](https://github.com/metafacture/metafacture-fix/pull/170/commits/e9dd4ab71b687e7cc9d9a570f4c5fe541762a69e#diff-dffdf77002e060149adb828b760328d625cd5cf1ba714c7020038049cc6090e4). The handling here is required for operations working on the record after such deletions (covered by `shouldDeleteNonFoundLookupOnDemandAndVacuum`).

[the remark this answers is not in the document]
blackwinter: I see, thanks. So there's a `new Value((String) null)` inserted.

In that case, I would suggest something along those lines:

```diff
diff --git metafix/src/main/java/org/metafacture/metafix/Record.java metafix/src/main/java/org/metafacture/metafix/Record.java
index 852a22f..1c1180e 100644
--- metafix/src/main/java/org/metafacture/metafix/Record.java
+++ metafix/src/main/java/org/metafacture/metafix/Record.java
@@ -201,13 +201,18 @@ public class Record extends Value.Hash {
         final Value found = findPath.findIn(this);
         findPath.throwIfNonString(found);
         Value.asList(found, results -> {
-            for (int i = 0; i < results.size(); ++i) {
+            for (int i = results.size() - 1; i >= 0; --i) {
                 final Value oldValue = results.get(i);
                 final FixPath insertPath = findPath.to(oldValue, i);
                 oldValue.matchType()
                     .ifString(s -> {
-                        final Value newValue = new Value(operator.apply(s));
-                        insertPath.insertInto(this, InsertMode.REPLACE, newValue);
+                        final String newValue = operator.apply(s);
+                        if (newValue == null) {
+                            insertPath.removeNestedFrom(this);
+                        }
+                        else {
+                            insertPath.insertInto(this, InsertMode.REPLACE, new Value(newValue));
+                        }
                     })
                     .orElseThrow();
             }
diff --git metafix/src/main/java/org/metafacture/metafix/Value.java metafix/src/main/java/org/metafacture/metafix/Value.java
index 7d30575..9c5b579 100644
--- metafix/src/main/java/org/metafacture/metafix/Value.java
+++ metafix/src/main/java/org/metafacture/metafix/Value.java
@@ -325,7 +325,7 @@ public class Value {
                 })
                 // TODO: Catmandu considers whitespace-only strings empty (`$v !~ /\S/`)
                 .ifString(s -> c.accept(s.isEmpty()))
-                .orElse(w -> c.accept(true))
+                .orElseThrow()
         );
 
         @Override
```

Then `lookup/fromJson/toJson/lookupSimpleWithInternalMapDeleteNotMatchingLiteral` also passes.

blackwinter:
> 2. `uniq()` does not work anymore due to the inclusion of `path` in `Value.equals()`.

This was resolved with 34c96f1.

> 1. It's unclear where they're coming from but we get `null` values in records now: `Error while executing Fix expression [...]: vacuum()`, caused by `java.lang.IllegalStateException: Expected Array or Hash or String, got null`.
> 3. `lookup()` behaviour seems to have changed unexpectedly (see also #149).

These would be resolved with my [suggestion above](https://github.com/metafacture/metafacture-fix/pull/170#discussion_r823631287).

> 4. Entity boundaries are violated (probably due to some error/change w.r.t. `$append`).

This seems to be related to `Record.transform()`, not `$append`. See 63bf98a for a failing test case.

metafix/src/test/resources/org/metafacture/metafix/integration/lookup/fromJson/toJson/lookupSimpleWithInternalMapDeleteNotMatchingLiteral/todo.txt
```diff
@@ -0,0 +1 @@
+See https://github.com/metafacture/metafacture-fix/issues/149
```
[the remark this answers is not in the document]
blackwinter: Would be resolved with my [suggestion above](https://github.com/metafacture/metafacture-fix/pull/170#discussion_r823631287). Although I don't have an explanation yet.

metafix/src/main/java/org/metafacture/metafix/Value.java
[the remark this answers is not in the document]
fsteeg: Great, thanks, pushed in [238e840](https://github.com/metafacture/metafacture-fix/pull/170/commits/238e8406e9868487201317a0f9a68ce98fba21a3).

metafix/src/test/resources/org/metafacture/metafix/integration/lookup/fromJson/toJson/lookupSimpleWithInternalMapDeleteNotMatchingLiteral/todo.txt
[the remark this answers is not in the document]
fsteeg: Found the difference: the integration test has non-repeated, single fields. I've added a corresponding unit test in [006e3ac](https://github.com/metafacture/metafacture-fix/pull/170/commits/006e3acf54e3584ba4e4303f99c73e4f9adc9c05#diff-a99cbcfd433b15909a7b85a7a04e6c5f4700ee0d33b2c016de2a2f15138c06ff).

fsteeg:
> 4. Entity boundaries are violated [...] See https://github.com/metafacture/metafacture-fix/commit/63bf98a09c26444c0c070c42a02def5637b68691 for a failing test case.

Thanks for adding the test. Fixed with [9e55ff3](https://github.com/metafacture/metafacture-fix/pull/170/commits/9e55ff3598831305a46120890d5f7888eae569b4) (which also removes the `*` workarounds in `Path#to`).

metafix/src/main/java/org/metafacture/metafix/FixPath.java
```diff
@@ -100,124 +107,59 @@ private Value findInValue(final Value value, final String[] p) {
...
-                    }
-                })
-                .orElseThrow();
-        }
-    }
-
     @Override
     public String toString() {
-        return Arrays.asList(path).toString();
+        return Joiner.on(".").join(Arrays.asList(path));
```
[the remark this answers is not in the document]
fsteeg: Oh, right, thanks, fixed in da719e7.

blackwinter:
Unfortunately, 9e55ff3 breaks `lookup()` in Limetrans.

fsteeg:
> Unfortunately, https://github.com/metafacture/metafacture-fix/commit/9e55ff3598831305a46120890d5f7888eae569b4 breaks lookup() in Limetrans (see https://github.com/metafacture/metafacture-fix/commit/ad0f3d54b76a37e4561f9dd8545ea2cb83ceff2f for a failing test case).

Thanks for the test case. Fixed in [33d103f](https://github.com/metafacture/metafacture-fix/pull/170/commits/33d103f94a32ec85a1286b07b83c50b9c37d8914) (plus corresponding test case for `move_field`).

fsteeg:
> Will try to find a better test. [...] See https://github.com/metafacture/metafacture-fix/commit/771205fc2c539d3e0332b79b6a294bb55983341c.

Thanks, fixed in [49849a4](https://github.com/metafacture/metafacture-fix/pull/170/commits/49849a4e35887aae4960f91d5b70c93ada14eda1).

blackwinter:
More failures in fa24c36.

fsteeg:
Thanks for the additional test cases, works with https://github.com/metafacture/metafacture-fix/pull/170/commits/d5e67fc21e0da624266d9f427dd52eda72670059.

blackwinter:
So here are some new failures in Limetrans:

```diff
   "RSWK" : [ {
-    "subjectTopicName" : "Building materials"
+    "subjectTopicName" : "Building materials.",
+    "a" : "Building materials."
   } ]
```

```diff
   "RSWK" : [ {
-    "subjectGenre" : "Periodicals",
-    "subjectTopicName" : "Art"
-  }, {
-    "subjectIdentifier" : [ "(OCoLC)fst00815326" ]
+    "subjectTopicName" : "Art",
+    "subjectIdentifier" : [ "(OCoLC)fst00815326" ],
+    "a" : "Art"
   } ]
```

```diff
   "OnlineAccess" : [ {
+    "uri" : "http://www.bibliothek.uni-regensburg.de/ezeit/?2241653",
     "nonpublicnote" : "EZB",
-    "uri" : "http://www.bibliothek.uni-regensburg.de/ezeit/?2241653"
+    "u" : "http://www.bibliothek.uni-regensburg.de/ezeit/?2241653"
   }, {
```

```diff
   "OnlineAccess" : [ {
     "publicnote" : "De Gruyter Basics Online",
+    "uri" : "https://eu04.alma.exlibrisgroup.com/view/uresolver/49HBZ_WUP/openurl?u.ignore_date_coverage=true&portfolio_pid=5354967650006447&Force_direct=true",
     "isInternalLinkFor" : "DE-468",
-    "uri" : "https://eu04.alma.exlibrisgroup.com/view/uresolver/49HBZ_WUP/openurl?u.ignore_date_coverage=true&portfolio_pid=5354967650006447&Force_direct=true"
+    "$last" : "https://eu04.alma.exlibrisgroup.com/view/uresolver/49HBZ_WUP/openurl?u.ignore_date_coverage=true&portfolio_pid=5354967650006447&Force_direct=true"
   } ],
```

In addition to a new exception `Expected Hash or Array, got String` for `lookup("@current_library.*", "alma-library-to-branch-$[isil]", delete: "true")`.

I have no idea what to make of it. I'm really sorry, but the band-aids are getting way out of hand, don't you think? Wouldn't it be more prudent at this point to scrap 9e55ff3 and call it a day? (Keep additional tests and any (independent) fixes, though.)

That would only leave us with the remaining [issue](https://github.com/metafacture/metafacture-fix/pull/170#discussion_r820674213) of `FixPath.throwIfNonString()`, but I'd be willing to concede it if that means we can move on.

WDYT?

fsteeg:
> Wouldn't it be more prudent at this point to scrap 9e55ff3 and call it a day? (Keep additional tests and any (independent) fixes, though.)

Alright, sounds good. I've reset to 44dc0be (parent of 9e55ff3) and added 5494f42 & 1e8ac6e.

blackwinter:
Okay, thanks. AFAICS, the only remaining issue in Limetrans is 63bf98a; everything else can be worked around. Once that's fixed (without introducing new problems), we should be able to merge.

fsteeg:
> AFAICS, the only remaining issue in Limetrans is https://github.com/metafacture/metafacture-fix/commit/63bf98a09c26444c0c070c42a02def5637b68691 [...]

Pushed 64d8d989c4ce805a8343e55b6bdef7237d41c985 to fix 63bf98a09c26444c0c070c42a02def5637b68691, hopefully without breaking other cases.

blackwinter:
It seems there's still a different issue: `FixPath.removeNestedFrom(Array)` removes the value at an index even if the path has not been fully exhausted (see 8cfe1a8 for a test).

fsteeg:
Thanks for the test case and analysis! e2cf74c779327d77f93e99a1710391668cdc063b seems to do the trick and should make sense too.

blackwinter:
Now if only the order would be preserved ;)

See 2378455:

- `shouldPreserveOrderWhenTransformingArraySubfields` produces 3, 2, 1 instead of 1, 2, 3.
- `shouldPreserveOrderWhenTransformingOptionalArraySubfield` produces `transformed value`, `other value` instead of `null`, `transformed value`.

fsteeg:
Alright, here is my suggestion: f210dd3 preserves the order, but breaks the behavior on faulty lookup usage (marked as TODO, does not break any intended behavior, 'just' error handling) and a4d557e manually sets the path in the test, as it happens automatically in Metafix.java (not ideal, but should work for usage as in the test; if needed, an alternative could be to automatically set the paths recusively starting in Record#add).

blackwinter:
> f210dd3 [...] breaks the behavior on faulty lookup usage (marked as TODO, does not break any intended behavior, 'just' error handling)

Sorry, but no, this definitely breaks intended behaviour:

```text
    org.metafacture.metafix.FixExecutionException: Error while executing Fix expression (at file:[...]/limetrans/build/resources/main/transformation/alma.fix, line 1910): lookup("@language_long.*", "ISO639-2-to-GND", delete: "true")
[...]
    Caused by: java.lang.IndexOutOfBoundsException: Index 1 out of bounds for length 1
        at jdk.internal.util.Preconditions.outOfBounds(Preconditions.java:64) ~[?:?]
        at jdk.internal.util.Preconditions.outOfBoundsCheckIndex(Preconditions.java:70) ~[?:?]
        at jdk.internal.util.Preconditions.checkIndex(Preconditions.java:266) ~[?:?]
        at java.util.Objects.checkIndex(Objects.java:359) ~[?:?]
        at java.util.ArrayList.set(ArrayList.java:441) ~[?:?]
        at org.metafacture.metafix.Value$Array.set(Value.java:414) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.FixPath$InsertMode$1.apply(FixPath.java:174) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.FixPath.insertInto(FixPath.java:248) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.FixPath.lambda$insertInto$20(FixPath.java:285) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Value$TypeMatcher.match(Value.java:302) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Value$TypeMatcher.ifArray(Value.java:275) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.FixPath.insertInto(FixPath.java:285) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Record.lambda$null$1(Record.java:214) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Value$TypeMatcher.match(Value.java:302) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Value$TypeMatcher.ifString(Value.java:283) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Record.lambda$transform$2(Record.java:208) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Value.asList(Value.java:190) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Value.asList(Value.java:184) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.Record.transform(Record.java:203) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.FixMethod$31.apply(FixMethod.java:386) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.RecordTransformer.lambda$processFunction$4(RecordTransformer.java:163) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.RecordTransformer.lambda$processExpression$6(RecordTransformer.java:172) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        at org.metafacture.metafix.RecordTransformer.processFix(RecordTransformer.java:177) ~[metafix-0.2.0-SNAPSHOT.jar:?]
        ... 92 more
```

I have no idea why this isn't covered by any tests...

metafix/src/main/java/org/metafacture/metafix/Record.java
```diff
@@ -153,4 +190,52 @@ public void retainFields(final Collection<String> fields) {
...
+     *
+     * @param field The field
+     * @param operator The operator
+     */
+    public void transform(final String field, final UnaryOperator<String> operator) {
+        final FixPath findPath = new FixPath(field);
+        final Value found = findPath.findIn(this);
+        findPath.throwIfNonString(found);
+        Value.asList(found, results -> {
+            for (int i = 0; i < results.size(); ++i) {
```
blackwinter, line 204: You can't iterate in ascending order if you're going to [remove](https://github.com/metafacture/metafacture-fix/pull/170#discussion_r823631287) elements based on the index number - they're going to be off once the first element has been removed.

blackwinter: Here's a test for the [above mentioned](https://github.com/metafacture/metafacture-fix/pull/170#issuecomment-1081983278) error (introduced by f210dd3): a816a73. It's exactly the issue described here.

metafix/src/test/java/org/metafacture/metafix/RecordTest.java
```diff
@@ -343,20 +341,20 @@ public void shouldPreserveOrderWhenTransformingArraySubfields() {
...
+        final String path = String.join(".", FIELD, "%s", OTHER_FIELD);
         record.put(FIELD, Value.newArray(a -> {
             a.add(Value.newHash(h -> {
                 h.put(FIELD, VALUE);
             }));
             a.add(Value.newHash(h -> {
                 h.put(FIELD, VALUE);
-                h.put(OTHER_FIELD, OTHER_VALUE);
+                // For proper transformation, we need to explicitly set the path (as in Metafix.java):
+                h.put(OTHER_FIELD, new Value(OTHER_VALUE.asString(), String.format(path, a.size() + 1)));
```
blackwinter, line 354: Now I have to admit that I'm a little concerned by this. Shouldn't this be done by `Value.Array.add()` then? I don't really think it's reasonable to expect users of the API to set the paths appropriately. Especially if it's only rarely actually required.

metafix/src/main/java/org/metafacture/metafix/Record.java
fsteeg: Right, thanks for the details and test case. Fixed with ec9f8ef.

metafix/src/test/java/org/metafacture/metafix/RecordTest.java
fsteeg: I agree, and as mentioned above, I currently think the best place would be in Record#add, since we have the full path information available there. If we want to set it in Array, I think we'd have to track/append the path (as I tried before). But it seems this is no regression as the test doesn't pass in master (unless I made some mistake while adapting it for master), so I'd suggest we call this a month and shift remaining problems to smaller issues / PRs.

blackwinter:
> I'd suggest we call this a month and shift remaining problems to smaller issues / PRs.

No such luck. There's still (at least) one regression in Limetrans. I'll try to come up with a unit test, again.

blackwinter:
Transformation of sub fields in an array works with actual arrays, bot not with repeated fields; see 725097a.

fsteeg:
> Transformation of sub fields in an array works with actual arrays, but not with repeated fields; see https://github.com/metafacture/metafacture-fix/commit/725097a4d1daa2e874046cff8f76fce7299ae5d2.

This one is solved with 4324378.

blackwinter:
> This one is solved with 4324378.

I'm sorry, but the unit test doesn't seem to capture the essence of the issue again. So here's an integration test instead: ff6fc72. Can you spot the difference?

blackwinter:
That's better, thanks. But not quite there yet: 82bee32.

fsteeg:
Let's see if this helps: 049c709.

metafix/src/main/java/org/metafacture/metafix/FixPath.java
```diff
@@ -116,6 +116,17 @@ public String toString() {
     // try to replace this with consistent usage of Value#getPath
     // (e.g. take care of handling repeated fields and their paths)
 
+    /*package-private*/ void throwIfNonString(final Value value) {
+        final boolean isNonString = value != null &&
+            // basic idea: path is only set on literals/strings
+            value.getPath() == null &&
+            // but with wildcards, we might still point to literals/strings
+            !hasWildcard();
```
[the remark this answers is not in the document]
blackwinter: Added a test in 96ed9b3.

blackwinter:
> I'll do a more thorough check tomorrow.

Larger data set didn't uncover any new issues.

Except maybe performance: This branch slows down our transformation of about 3500 MARCXML records by about 15-20 % (from ~7m to ~8m15s).

blackwinter:
> Tweaked MetafixScriptTest for order issue in 4961b95.

The behaviour has changed and that should be reflected in the test: 0dfcfb8.

> The remaining issue in MetafixScriptTest seems to have the same cause as 7b74937 (not throwing on transform of non-string field).

No, both `shouldSkipExpressionOnExecutionException` and `shouldSkipRecordOnExecutionException` are throwing an exception, but should react differently.

metafix/src/test/java/org/metafacture/metafix/MetafixScriptTest.java
```diff
@@ -430,41 +429,41 @@ private void assertStrictness(final Metafix.Strictness strictness, final boolean
     }
 
     @Test
-    @MetafixToDo("See https://github.com/metafacture/metafacture-fix/pull/170")
```
fsteeg: But it's not working, right? It dosen't skip any expression, because there is no exception.

metafix/src/test/java/org/metafacture/metafix/MetafixScriptTest.java
```diff
@@ -430,41 +429,41 @@ private void assertStrictness(final Metafix.Strictness strictness, final boolean
...
     public void shouldSkipExpressionOnExecutionException() {
         assertStrictness(Metafix.Strictness.EXPRESSION, true, o -> {
             o.get().startRecord("1");
-            o.get().literal("before", "");
             o.get().literal("data", "FOO");
+            o.get().literal("before", "");
```
fsteeg, line 436: Then I think it would be better to call the fields "one", "two" (not "before", "after").

metafix/src/test/java/org/metafacture/metafix/MetafixScriptTest.java
blackwinter: Yes, it does throw an exception (change `Metafix.Strictness.EXPRESSION` to `Metafix.Strictness.PROCESS` to see for yourself; that's what the other tests are confirming as well) and it skips the `upcase()` expression (values remain lowercase).

metafix/src/test/java/org/metafacture/metafix/MetafixScriptTest.java
blackwinter: But sure, call it whatever makes (more) sense to you.

metafix/src/test/java/org/metafacture/metafix/MetafixScriptTest.java
fsteeg: Oh, so leaving the values lowercase, but including them, is the desired behavior?

fsteeg: Right, well I do think it makes sense to leave the values untouched if something goes wrong, instead of removing them. I was just surprised because that was not the behavior of the original test (which passed in master, and I marked as TODO when merging). So +1 from me here.

blackwinter: Agreed. This is also what Catmandu does (hbz/Catmandu@370ca93).

As for the cause: Previously, the field would be [removed](https://github.com/metafacture/metafacture-fix/blob/c25517aaa2958b0ae98420b0f83c9962f176c944/metafix/src/main/java/org/metafacture/metafix/FixPath.java#L149) before the value's type was [checked](https://github.com/metafacture/metafacture-fix/blob/c25517aaa2958b0ae98420b0f83c9962f176c944/metafix/src/main/java/org/metafacture/metafix/FixPath.java#L154); now the [check](https://github.com/metafacture/metafacture-fix/blob/0dfcfb86f725baa73989e6fac52018f9f7c4049e/metafix/src/main/java/org/metafacture/metafix/Record.java#L204) happens before the [removal](https://github.com/metafacture/metafacture-fix/blob/0dfcfb86f725baa73989e6fac52018f9f7c4049e/metafix/src/main/java/org/metafacture/metafix/Record.java#L220), so an exception due to an unexpected type prevents the value from being removed.

metafix/src/main/java/org/metafacture/metafix/FixPath.java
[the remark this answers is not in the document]
fsteeg: Pushed 2b476e2 to address this.

metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
```diff
@@ -305,10 +305,32 @@ public void shouldNotTrimRepeatedField() {
...
+                }
+            )
+        );
+    }
+
     @Test
     // See https://github.com/metafacture/metafacture-fix/pull/133
     public void shouldNotTrimStringInImplicitArrayOfHashes() {
-        MetafixTestHelpers.assertExecutionException(IllegalStateException.class, "Expected String, got Array", () ->
+        MetafixTestHelpers.assertExecutionException(NumberFormatException.class, "For input string: \"title\"", () ->
```
blackwinter, line 333: Hm, this doesn't look right. I think it should still be an `IllegalStateException`. Maybe "Expected Hash, got Array" then.

metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
```diff
@@ -305,10 +305,32 @@ public void shouldNotTrimRepeatedField() {
         );
     }
 
+    @Test
+    public void shouldNotTrimIndexedArray() {
```
blackwinter, line 309: It's not an "indexed array", is it? The entity is missing the array marker.

metafix/src/main/java/org/metafacture/metafix/FixPath.java
```diff
@@ -137,7 +148,7 @@ else if (value.getPath() != null && (countAsterisks() >= 2) || hasWildcard()) {
     }
 
     private boolean hasWildcard() {
-        return Arrays.asList(path).stream().filter(s -> s.contains("?") || s.contains("|") || s.matches(".*?\\[.+?\\].*?")).findAny().isPresent();
+        return Arrays.asList(path).stream().filter(s -> s.equals("*") || s.contains("?") || s.contains("|") || s.matches(".*?\\[.+?\\].*?")).findAny().isPresent();
```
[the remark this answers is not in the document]
fsteeg: Improved this a bit with 6779544.

metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
fsteeg: Yes, I agree, addressed in 746fcf4.

metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
fsteeg: Right, thanks, tweaked and added test in 9b8fa39.

metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
blackwinter: Thanks. I guess I would've placed it in `FixPath.InsertMode.REPLACE.apply(Array, String, Value)`. But doesn't matter too much at this point.

fsteeg: Right, that's better. Pushed 5a3fa9b.

blackwinter (approved):
The moment of truth, huh? :worried: :fearful: :cold_sweat:

(`@MetafixToDo`: 23 -> 29, `todo.txt`: 86 -> 72)

fsteeg:
> blackwinter approved these changes

\o/

> (`@MetafixToDo`: 23 -> 29, `todo.txt`: 86 -> 72)

But we fixed the regressions, these are additional tests. More tests (even failing) are good, no?

Looking at passing tests, I'm very happy: 353 -> 401 (unit), 141 -> 155 (integration).

blackwinter:
> But we fixed the regressions, these are additional tests. More tests (even failing) are good, no?

Yes, I was just tying into the previous remarks from way back when this discussion started ;)

blackwinter:
Found another regression after all :(

Tests and (partial) fix in 1ecbcdf.

Issue #148, opened 2022-02-14
array of strings + copy_field with $append


TobiasNx:
```
{
    "a": ["b", "c", "d"],
    "w": "e",
    "z": []
}
```

fix:
```
copy_field("a[]", "z[].$append")
copy_field("a[].*", "z[].$append")
copy_field("w", "z[].$append")
```

result:
```
{
  "a" : [ "b", "c", "d" ],
  "w" : "e",
  "z" : [ "b", "c", "d", "b", "c", "d", "e" ]
}
```

expected:
```
{
  "a" : [ "b", "c", "d" ],
  "w" : "e",
  "z" : [ ["b", "c", "d"], "b", "c", "d", "e" ]
}
```

`copy_field("a[]", "z[].$append")` without `*` operates exactly as  `copy_field("a[].*", "z[].$append")`

This is in my opinion wrong. I will build a test for that later.

fsteeg:
Revisit / add as unit test after https://github.com/metafacture/metafacture-fix/issues/102.

blackwinter:
Integration tests:

- [copy_fieldValuesOfArrayOfStringsIntoArray](https://github.com/metafacture/metafacture-fix/tree/master/metafix/src/test/resources/org/metafacture/metafix/integration/record/fromJson/toJson/copy_fieldValuesOfArrayOfStringsIntoArray)
- [move_fieldWithAlternationOfFieldInArrayOfObjects](https://github.com/metafacture/metafacture-fix/tree/master/metafix/src/test/resources/org/metafacture/metafix/integration/record/fromJson/toJson/move_fieldWithAlternationOfFieldInArrayOfObjects)

PR #175, opened 2022-02-28
Add integration test for #148

TobiasNx:
Integration test for #148


blackwinter:
Why did you re-request review without addressing prior feedback?

TobiasNx:
It should be now

Issue #132, opened 2022-02-04
`copy_field` of array/repeated field into array with $first/$last creates worng output


TobiasNx:
Spotted while testing #106: https://github.com/TobiasNx/fix-FunctionalReview-Testing/blob/dfcf4fac13e4deab0606a9131ea846175170cfbc/data/testing/repeatedFieldsArrayToXML

Perhaps this is also a problem for `add_field` and `move_field`.
Should be dealt with after: #102

```
copy_field("animal_string_Array[]", "animals_repeated_SimpleField")
copy_field("animal_object_Array[]", "animals_repeated_ObjectField")
copy_field("animal_object_Array[]", "test_animals_repeated_ObjectField")
copy_field("animal_object_Array[]", "test_animal_object_Array[]")

#This creates trouble - writes into every first object in array
copy_field("animal_string_Array[]", "test_animals_repeated_ObjectField.$first.animalsString_1")
#This creates trouble - writes into every last object in array
copy_field("animal_string_Array[]", "test_animals_repeated_ObjectField.$last.animalsObject_1[]")

#This creates trouble - writes into every first object in array
copy_field("animal_string_Array[]", "test_animal_object_Array[].$first.animalsString_2")
#This creates trouble - writes into every last object in array
copy_field("animal_string_Array[]", "test_animal_object_Array[].$last.animalsObject_2[]")
```

expected:
```
{
  "animal_string_Array" : [ "dog", "elefant" ],
  "animal_object_Array" : [ {
    "name" : "dog"
  }, {
    "name" : "elefant"
  } ],
  "animals_repeated_SimpleField" : "dog",
  "animals_repeated_SimpleField" : "elefant",
  "animals_repeated_ObjectField" : {
    "name" : "dog"
  },
  "animals_repeated_ObjectField" : {
    "name" : "elefant"
  },
  "test_animals_repeated_ObjectField" : {
    "name" : "dog",
    "animalsString_1" : "dog",
    "animalsString_1" : "elefant"
  },
  "test_animals_repeated_ObjectField" : {
    "name" : "elefant",
    "animalsObject_1" : [ "dog", "elefant" ]
  },
  "test_animal_object_Array" : [ {
    "name" : "dog",
    "animalsString_2" : "dog",
    "animalsString_2" : "elefant"
  }, {
    "name" : "elefant",
    "animalsObject_2" : [ "dog", "elefant" ]
  } ]
}
```

```
{
[...]

Issue #121, opened 2022-02-01
Inconsistent behaviour with asterisk-wildcard in array path


TobiasNx:
Since we do not have a ticket for that here. The asterisk wildcard in array paths seems to work sometimes, sometimes not and sometimes it breaks. I have set up a test with multiple usages of the `*`-wildcard and separated between working, breaking and no-doers. I am not sure if this is due to inconsistent array handling or due to the function. But I think it is good to collect them here:

https://github.com/TobiasNx/fix-FunctionalReview-Testing/blob/a8f314ce0c416073873553166eb9116e41a95cb7/data/testing/asteriskArray

Following functions work, also with arrays in arrays:
`prepend`
`append`
`replace`

`copy_field` works with single asterisk but not with arrays in arrays.

Following do nothing:
`split_field`
`sort_field`
`add_field`

Following break with asterisk:
`reverse`
`sum`
`rename`

blackwinter:
There's a pattern w.r.t. how they're implemented internally:

- `append`, `prepend`, `replace_all`: `Record.transformFields()`
- `rename`, `reverse`, `sort_field`, `split_field`, `sum`: `Record.transformField()` [no `s`]
- `add_field`: `Record.append()`
- `copy_field`: `Record.copy()`

blackwinter:
> `copy_field` works with single asterisk but not with arrays in arrays.

You mean it should work without `set_array(...)`? Otherwise, I don't see a difference between

```
set_array("TEST_TWO[]")
copy_field("test[].*", "TEST_TWO[].$append")
```

and 

```
set_array("TEST_3[]") # sic!
copy_field("test[].*", "TEST_THREE[].$append")
```

blackwinter:
Oh, I guess you meant

```
set_array("TEST_4[]")
copy_field("nestedTest[].*.test[].*", "TEST_4[].$append")
```

So this is the _expected_ output for `TEST_THREE` then?

```json
"TEST_THREE" : [ {
  "1" : "One",
  "2" : "Two",
  "3" : "Three"
} ]
```

TobiasNx:
No actually this is an error in my fix. I corrected it TEST_3 is connected to the reverse handling that is commented out:
https://github.com/TobiasNx/fix-FunctionalReview-Testing/blob/764464e662ad0cae2b176e3b6087784709dde8b9/data/testing/asteriskArray/asteriskArray.fix#L27-L28

TobiasNx:
[...]

PR #180, opened 2022-03-01
Add integration test for #132

TobiasNx:
Integration Tests for #132
Mostly control , it seems that the problem has something to do with created arrays.

Also one for #121


metafix/src/test/resources/org/metafacture/metafix/integration/record/fromJson/toJson/copy_fieldArrayOfObjectAsRepeatedObjectAndThenArrayOfStringsIntoTheArrayOfObjectsWithLastWildcard/test.fix
```diff
@@ -0,0 +1,2 @@
+copy_field("test[]", "repeatedTest")
```
blackwinter, line 1: Where is this field in `expected.json`?

metafix/src/test/resources/org/metafacture/metafix/integration/record/fromJson/toJson/input_old.json
```diff
@@ -0,0 +1,10 @@
+{
```
blackwinter, line 1: What does this file do? Please move or remove it.

blackwinter (changes requested):

blackwinter (approved):

PR #184, opened 2022-03-02
160-addIntegrationTestsFor121

TobiasNx:
Integration tests for #121 and functions that are discussed in that issue


metafix/src/test/resources/org/metafacture/metafix/integration/method/fromJson/toJson/sumSubfieldInArrayOfObjects/input.json
```diff
@@ -0,0 +1,7 @@
+{
+  "objects" : [ {
+    "numbers" : [ "7", "2", "7", "1", "10" ]
+  }, {
+    "numbers_2" : [ "1", "2", "3", "4"]
```
blackwinter, line 5: This should be `numbers`.

blackwinter (changes requested):
- `rename()` should be in the `record` area.
- `ArrayOfObject` should be `ArrayOfObjects`.

blackwinter (approved):

PR #194, opened 2022-03-04
160-moreIntegrationTests

TobiasNx:
A lot more integration tests


blackwinter:
Test names should start with lowercase (`If` -> `if`).

blackwinter (changes requested):
Typo: `Refrence` -> `Reference`

Expected files must be formatted correctly; I've given up marking them individually...

metafix/src/test/resources/org/metafacture/metafix/integration/record/fromJson/toJson/set_arrayNestedSimple/test.fix
```diff
@@ -0,0 +1,2 @@
+set_array("object.test[]")
+set_array("object_2.test_2[]", "dog", "cat", "bird")
```
blackwinter, line 2: Should probably be two tests.

blackwinter (changes requested):
Typo still present: `Refrence` -> `Reference`

metafix/src/test/resources/org/metafacture/metafix/integration/record/fromJson/toJson/set_arrayNestedSimple/test.fix
TobiasNx: I have split the `set_array` tests. Found an error and added some more variations of the tests.

blackwinter (approved):

PR #198, opened 2022-03-11
Add infrastructure for profiling and benchmarking.

blackwinter:
Basic infrastructure for integrating profiling ([HPROF](https://docs.oracle.com/javase/8/docs/technotes/samples/hprof.html)) and benchmarking ([JMH](https://github.com/openjdk/jmh)).

## Profiling

Optional heap and CPU profiling when running Flux.
    
- Specify `-Pprofile[=<filename>]` to enable profiling for `:metafix-runner:run`.
- Specify `-Pprofile=true|false` to enable/disable profiling for `:metafix:integrationTest`.
- Specify `METAFIX_INTEGRATION_TEST_PROFILE=true|false` to enable/disable profiling for `metafix/integrationTest.sh`.

### Result

Integration test `script/fromJson/toJson/nothing`:

```
SITES BEGIN (ordered by live bytes) Fri Mar 11 14:15:18 2022
          percent          live          alloc'ed  stack class
 rank   self  accum     bytes objs     bytes  objs trace name
    1 33.97% 33.97%  33554448    1  33554448     1 301112 char[]
    2  8.49% 42.46%   8388624    1   8388624     1 314082 byte[]
    3  6.29% 48.75%   6209032 1468  24573952  6141 300385 byte[]
    4  4.15% 52.90%   4096968  820   7124408  1639 304687 byte[]
    5  2.67% 55.57%   2638304 13128  11576008 61279 300166 byte[]
    6  2.65% 58.22%   2617824 9553   8097776 24269 300010 char[]
    7  1.86% 60.08%   1837200 13144   8084488 61308 300165 char[]
    8  1.14% 61.22%   1125616 8541   2332960 16998 301238 char[]
    9  1.09% 62.31%   1078016 5999   1422680  7964 301248 char[]
   10  1.06% 63.37%   1051624 13144   4653112 61308 300167 byte[]
...
```

```
CPU SAMPLES BEGIN (total = 1062) Fri Mar 11 14:15:18 2022
rank   self  accum   count trace method
   1  4.61%  4.61%      49 300520 java.lang.ClassLoader.defineClass1
   2  2.35%  6.97%      25 300165 java.lang.String.toCharArray
   3  2.35%  9.32%      25 300166 java.util.zip.ZipCoder.getBytes
   4  2.35% 11.68%      25 300167 java.util.Arrays.copyOf
   5  2.07% 13.75%      22 301203 java.lang.Object.<init>
   6  1.79% 15.54%      19 301290 java.lang.Object.<init>
[...]

PR #209, opened 2022-03-29
Fix missing resource _UI_DiagnosticRoot_diagnostic

dr0i:
This silences the ERROR message when doing "./gradlew jmh".

See https://github.com/metafacture/metafacture-fix/pull/198.


blackwinter (changes requested):
Awesome, thanks! But I'd prefer the file to live in the `jmh` source directory (`metafix/src/jmh/resources/plugin.properties`).

How did you arrive at this solution? What does this change do under "normal" circumstances, i.e. other tasks/contexts than `jmh`?

dr0i:
> How did you arrive at this solution? 

Just [googled](https://www.startpage.com/sp/search?q=_UI_DiagnosticRoot_diagnostic). 

> What does this change do under "normal" circumstances, i.e. other tasks/contexts than jmh?

AFAIK this is just a very stupid default value which is sometimes needed. Also it is improperly logged as ERROR. Was not able to find the source code, though :/
Sometimes, when bundling jars, if there are multiple properties they overwrite themselves.
This commit just guarantees the existence of the property.
I don't expect any sideeffects.  Code was running good before that property was added and behaves the same after the adding of the property. Was just noise IMO (I understand your scepticism, but sometimes things are just stupid).

PR #218, opened 2022-05-06
Fix issue with dollars in var paths when handled as regex

fsteeg:
Found via regression in OERSI, see new unit test and https://gitlab.com/oersi/oersi-etl/-/merge_requests/175

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Value.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Value.java
@@ -17,6 +17,7 @@
 package org.metafacture.metafix;
 
 import org.metafacture.commons.tries.SimpleRegexTrie;
+import org.metafacture.metafix.FixPath.ReservedField;
 
 import java.util.ArrayList;
 import java.util.Collection;
@@ -208,7 +209,10 @@ public class Value {
 
     /*package-private*/ Value updatePathRename(final String newName) {
         if (path != null) {
-            path = newName.replaceAll("\\$[^.]+", split(path)[0]);
+            final String basePath = split(path)[0];
+            for (final ReservedField rf : ReservedField.values()) {
+                path = newName.replace(rf.name(), basePath);
+            }
         }
         return this;
     }
--- a/metafix/src/main/java/org/metafacture/metafix/FixPath.java
+++ b/metafix/src/main/java/org/metafacture/metafix/FixPath.java
@@ -298,7 +298,7 @@ import java.util.Map;
         return Arrays.copyOfRange(fields, 1, fields.length);
     }
 
-    private enum ReservedField {
+    /*package-private*/ enum ReservedField {
         $append, $first, $last;
 
         private static final Map<String, ReservedField> STRING_TO_ENUM = new HashMap<>();
--- a/metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
+++ b/metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
@@ -2032,6 +2032,33 @@ public class MetafixMethodTest {
     }
 
     @Test
+    public void shouldCopyBindVarWithDollarAfterLookup() {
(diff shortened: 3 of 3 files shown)


blackwinter (approved):

metafix/src/main/java/org/metafacture/metafix/Value.java
```diff
@@ -208,7 +209,10 @@ public Value asList(final Consumer<Array> consumer) {
 
     /*package-private*/ Value updatePathRename(final String newName) {
         if (path != null) {
-            path = newName.replaceAll("\\$[^.]+", split(path)[0]);
+            final String basePath = split(path)[0];
+            for (final ReservedField rf : ReservedField.values()) {
+                path = newName.replace(rf.name(), basePath);
```
blackwinter, line 214: This leaves `path` at the *last* `ReservedField` replacement!

blackwinter (changes requested):

metafix/src/main/java/org/metafacture/metafix/Value.java
blackwinter: What about defining a static pattern in `FixPath` instead?

```java
/*package-private*/ static final Pattern RESERVED_FIELD_PATTERN = Pattern.compile(String.format("(?:%s)",
            Arrays.stream(ReservedField.values()).map(f -> Pattern.quote(f.name())).collect(Collectors.joining("|"))));
```

Then we can do a simple replacement again:

```java
path = FixPath.RESERVED_FIELD_PATTERN.matcher(newName).replaceAll(split(path)[0]);
```

fsteeg: Oh right, thanks! I've pushed your suggestion in https://github.com/metafacture/metafacture-fix/pull/218/commits/03bb744ea6af24c029dd8372609a531758459914, however I was not able to come up with a unit test for the issue. While trying, I've added 4 failing tests in https://github.com/metafacture/metafacture-fix/pull/218/commits/73ef45917141342e687344b27f52b95eee04579b, two of which I could fix with an unrelated change in https://github.com/metafacture/metafacture-fix/pull/218/commits/1a3988350dd45b1f3d9a8f999dab79dbc16dab71.

blackwinter: > however I was not able to come up with a unit test for the issue.

That's because only `$last` seems to actually require `Value.updatePathRename()`!?

If you adapt `MetafixMethodTest.replaceAllWithWildcardAfterCopyFieldWithVarInSourceAndTarget()` (the test fixed by ca1cba1) to use `$first` instead of `$last`, it passes even with an "empty" `Value.updatePathRename()`. So, because `$last` was the only `ReservedField` to require a path update and it was also the _last_ value in the enum, it incidentally worked even with the wrong implementation.

blackwinter: You can also verify that behaviour by temporarily changing the order of `ReservedField` values.

blackwinter (approved):

metafix/src/main/java/org/metafacture/metafix/Value.java
fsteeg: > As I said, change the order of ReservedField values and you'll see.

I did try that, I swapped `$first` and `$last`, added the tweaked `replaceAllWithWildcardAfterCopyFieldWithVarInSourceAndTarget` as you described, reverted https://github.com/metafacture/metafacture-fix/commit/03bb744ea6af24c029dd8372609a531758459914, but everything was passing.

blackwinter: Oh, that was for illustrating that `$first` doesn't appear to require `Value.updatePathRename()` ;)

I meant swapping `$first` and `$last` and then only updating `Value.updatePathRename()` (both with and without 03bb744).

Issue #207, opened 2022-03-28
Identify performance bottlenecks


dr0i:
@blackwinter experienced a serious decline of performance when comparing a comprehensive (~ 20M documents) ETL workflow using `fix` in comparison to a similar `morph`. Factor was about ~~100~~ ~~34~~ between 34 and 100.
It's very likely that there are a few bottlenecks which we should able be get rid off once identified.
A common approach is to use a profiler to identify bottlenecks.

With #198 there is a possibility to get a Memory Snapshot (`hprof`) which can be analyzed e.g. further tools like e.g. [MAT](https://www.eclipse.org/mat/).
 Unfortunately the produced `hprofs` via #198 are somehow faulty and cannot be loaded into MAT (or the native jhat). It's not clear if this is caused by using `gradle` (e.g. in conjunction with a preliminary cut-off writing the whole `hprof`) or a bug in java 1.8 or configuration error (tried several ones - i.a. of course using binary format, and `heap=all` etc). However, to identify a bottleneck we have to use real workflows with (lots of) real data. I am confident that such a Memory Snapshot is computable by e.g. MAT.

blackwinter:
> Unfortunately the produced `hprofs` via #198 are somehow faulty and cannot be loaded into MAT (or the native jhat).

Note that the default configuration only tracks allocation sites (`heap=sites`) not allocated objects, so `jhat` doesn't have anything to analyze. With `heap=dump` and `format=b` (along with changing the file extension from `txt` to `bin` for consistency) everything works fine.

blackwinter:
> @blackwinter experienced a serious decline of performance when comparing a comprehensive (~ 20M documents) ETL workflow using `fix` in comparison to a similar `morph`. Factor was about 100.

I have a more realistic data point by now: When running our workflow as it's supposed to apply to all our Alma libraries soon, we get a factor of about 34 (20 GB input file with approx. 23.5m records, filtered for a single library with approx 1.6m records).

- Metamorph: 1h 51m (final RSS: 510 MB)
- Metafix: 62h 45m (final RSS: 968 MB)

dr0i:
> heap=dump and format=b

Did that, it's consumable by jhat but not with MAT. jhat does not help me. 
I do this:
```
[~/git/metafacture-fix/metafix]
[...]

PR #223, opened 2022-05-31
WIP using int and modulo (#207)

dr0i:
Tried to use a faster data structure (integer and modulo instead of `HashSet`). Not sure how much of performance improvement this gains (`./gradlew jmh` seems to give too vague computations, and a real test running some minutes also doesn't yield an exact impression, however I think this may gain (for me surprisingly only) between 1 and 3 percent?).
Also two tests fail, unsure how to fix them or if they really matter.

This is more to give an idea what could be improved without refactoring a lot.
It obfuscates the code (could be improved a bit). Also a `boolean` array or better a `BitSet` could be used (tried that but with no fast success).

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Value.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Value.java
@@ -51,12 +51,12 @@ public class Value {
     private final Hash hash;
     private final String string;
 
-    private final Type type;
+    private final int type;
 
     private String path;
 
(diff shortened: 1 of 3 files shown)


blackwinter:
Interesting approach, thank you! I'll look into it as soon as possible.

blackwinter:
I've taken your idea and implemented two variants in #224. Thanks!

Issue #226, opened 2022-06-15
`paste` outputs values even if all fields are provided


TobiasNx:
[See here](https://metafacture.org/playground/?flux=PG_DATA%0A%7Cas-lines%0A%7Cdecode-formeta%0A%7Cfix%0A%7Cencode-yaml%0A%7Cprint%0A%3B&fix=paste%28author%2C+b.v%2C+b.n%2C+%27~aus%27%2C+c%29%0Aretain%28author%29&data=1%7Ba%3A+Faust%2C+b+%7Bn%3A+Goethe%2C+v%3A+JW%7D%2C+c%3A+Weimar%7D%0A2%7Ba%3A+R%C3%A4uber%2C+b+%7Bn%3A+Schiller%2C+v%3A+F%7D%7D&active-editor=fix)

I assumed that `paste` only outputs if all elements are provided. But it always pasts when a string is added [even when non is provided](https://metafacture.org/playground/?flux=PG_DATA%0A%7Cas-lines%0A%7Cdecode-formeta%0A%7Cfix%0A%7Cencode-yaml%0A%7Cprint%0A%3B&fix=paste%28author%2C+b.v%2C+b.n%2C+%27~aus%27%2C+c%29%0Aretain%28author%29&data=1%7Ba%3A+Faust%2C+b+%7Bn%3A+Goethe%2C+v%3A+JW%7D%2C+c%3A+Weimar%7D%0A2%7Ba%3A+R%C3%A4uber%7D&active-editor=fix). 

Input:
```
1{a: Faust, b {n: Goethe, v: JW}, c: Weimar}
2{a: Räuber, b {n: Schiller, v: F}}
```

fix:
```
paste(author, b.v, b.n, '~aus', c)
retain(author)
```

output:
```
---
author: "JW Goethe aus Weimar"

---
author: "F Schiller aus"

```

expected:
```
---
author: "JW Goethe aus Weimar"

---

```

blackwinter:
You marked this as a bug, but I for one am not entirely sure that it is. What is your expectation based on?

TobiasNx:
okay, @blackwinter you are probably right this is no bug in the sense that Catmandu is behaving similarly:
input:
```
{
  "name": "Blacky",
  "type" : "bird"
}
{
  "name" : "Test",
  "test_1" : "test_1"
}
{
  "test_1" : "test_1",
  "test_2" : "test_2"
}
```
without additional string:
```
[...]

PR #227, opened 2022-06-17
Add integration test for #226

TobiasNx:
Add integration test for #226


blackwinter (commented):
You marked both integration tests as "todo". But is there even consent yet that #226 is actually a bug?

TobiasNx:
No there is no explicit consent yet. But `paste` outputs even an empty field if there is no element that is matching the function. I think this behaviour is not right.

blackwinter:
Okay, let's reach an agreement on the issue first, shall we? Then we can decide what to do about it.

TobiasNx:
I am closing this PR, since metafix `paste` has the same behaviour as catmandu `paste`

PR #232, opened 2022-06-22
Add `timestamp()` Fix function.

### LINKED ISSUE #230: Add `timestamp`-function
At the moment there is no timestamp function like in [morph](https://github.com/metafacture/metafacture-core/blob/master/metamorph/src/main/java/org/metafacture/metamorph/functions/Timestamp.java).

I need this for the Alma-Transformation in Fix.

blackwinter:
Resolves #230.

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,16 +312,16 @@ hash("foo")
 
 #### `timestamp`
 
-Adds field or overwrites field with current timestamp.
+Creates (or replaces) a field with the current timestamp.
 
 Options:
 
-  - `format` patterns as in [java.text.SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html)
-  - `timezone` as in [java.util.TimeZone](https://docs.oracle.com/javase/8/docs/api/java/util/TimeZone.html)  (Standard: UTC)
-  - `language` area codes as in [java.util.Locale](https://docs.oracle.com/javase/8/docs/api/java/util/Locale.html)
+- `format`: Date and time pattern as in [java.text.SimpleDateFormat](https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html) (Default: Unix timestamp).
+- `timezone`: Time zone as in [java.util.TimeZone](https://docs.oracle.com/javase/8/docs/api/java/util/TimeZone.html) (Default: UTC).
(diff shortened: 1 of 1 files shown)


metafix/src/main/java/org/metafacture/metafix/api/FixFunction.java
```diff
@@ -24,13 +24,20 @@
...
 import java.util.Set;
+import java.util.function.Consumer;
 import java.util.stream.Stream;
 
 @FunctionalInterface
 public interface FixFunction {
 
     void apply(Metafix metafix, Record record, List<String> params, Map<String, String> options);
 
+    default void getOption(final Map<String, String> options, final String key, final Consumer<String> consumer) {
```
fsteeg, line 35: Maybe something like `consumeOption` or `getAndConsumeOption` would be clearer? (`getOption` is fine with me though).

fsteeg (approved):

metafix/src/main/java/org/metafacture/metafix/api/FixFunction.java
blackwinter: Yeah, I wasn't really sure about it myself. Naming things is hard ;)

Originally, I had `withOption()`. But then I opted for alignment with the existing `getBoolean()`/`getInteger()` instead.

fsteeg: > Originally, I had `withOption()`.

I like that, reads good: `withOption ... setFormat`

blackwinter: Okay, I've changed the name in 710db71.

fsteeg: :+1:

blackwinter:
Yes, that's the default behaviour:

> By default it returns a unix timestamp.

— [org.metafacture.metamorph.functions.Timestamp](https://javadoc.io/doc/org.metafacture/metamorph/latest/org/metafacture/metamorph/functions/Timestamp.html)

TobiasNx:
+1 I tested it and added documentation in the `README.md`

Thanks

blackwinter:
Thanks for the documentation! I've tweaked it a little.

PR #235, opened 2022-06-29
Add `in()` Fix conditional.

### LINKED ISSUE #234: Support Catmandu `in`-conditional
https://librecat.org/Catmandu/#conditionals

The `in` conditional compares a value of a source field with a target field.
If they have identical values then the condition is true. 

> `in(path1,path2)`
> 
> True, when the values of the first path1 are contained in the values at the second path2. False otherwise.
> 
> For instance to check if two paths contain the same values type:
> 
> ```
> if in(my.title,your.title)
>   set_field(same,1)
> end
> ```
> 
> To check if a value in one path is contained in a list of an other path type:
> 
> ```
> if in(my.author,your.authors.*)
>    set_field(known_author,1)
> end
> ```

This helps to avoid duplicate values or object when transforming metadata. I would need this for the ALMA-FIX.

blackwinter:
Resolves #234.

### DIFF
--- a/metafix/src/test/resources/org/metafacture/metafix/integration/conditional/fromJson/toJson/ifInArrayOfStrings/expected.json
+++ b/metafix/src/test/resources/org/metafacture/metafix/integration/conditional/fromJson/toJson/ifInArrayOfStrings/expected.json
@@ -1,0 +1,8 @@
+{
+  "test" : "cat",
+  "animals" : [ "cat", "dog", "zebra" ]
+}
+{
+  "test" : "cat",
+  "animal" : [ "parrot", "dog", "zebra" ]
+}
--- a/metafix/src/test/resources/org/metafacture/metafix/integration/conditional/fromJson/toJson/ifInArrayOfStrings/expected.json
+++ b/metafix/src/test/resources/org/metafacture/metafix/integration/conditional/fromJson/toJson/ifInArrayOfStrings/expected.json
@@ -1,6 +1,7 @@
 {
   "test" : "cat",
-  "animals" : [ "cat", "dog", "zebra" ]
+  "animals" : [ "cat", "dog", "zebra" ],
+  "this" : "works"
 }
 {
   "test" : "cat",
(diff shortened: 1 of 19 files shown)


metafix/src/test/resources/org/metafacture/metafix/integration/conditional/fromJson/toJson/ifInArrayOfStrings/expected.json
```diff
@@ -0,0 +1,8 @@
+{
+  "test" : "cat",
+  "animals" : [ "cat", "dog", "zebra" ]
+}
+{
+  "test" : "cat",
+  "animal" : [ "parrot", "dog", "zebra" ]
```
blackwinter, line 7: Expected `"this" : "works"`.

metafix/src/test/resources/org/metafacture/metafix/integration/conditional/fromJson/toJson/ifInArrayOfStrings/test.fix
```diff
@@ -0,0 +1,3 @@
+unless in("test","animals[]")
```
blackwinter, line 1: Doesn't match test name (`if` != `unless`).

fsteeg (approved):
(If we were building this from scratch, and not implementing a Catmandu feature, I think the parameters should be swapped. @TobiasNx Did we start collecting things we'd like to revisit when looking into standardizing the Fix language?)

blackwinter:
> If we were building this from scratch, and not implementing a Catmandu feature, I think the parameters should be swapped.

It checks if A is contained in B (in contrast to `*_contain`, which checks if A contains B), so I guess the order is fine as it is?

> It is often best read aloud as "in", "inside of", or "is contained in", because the left operand is often looked for _inside_ the right operand. That makes the order of the operands to the smartmatch operand often opposite that of the regular match operator. In other words, the "smaller" thing is usually placed in the left operand and the larger one in the right.

— [Perl Smartmatch Operator](https://perldoc.perl.org/perlop#Smartmatch-Operator)

fsteeg:
But with the operator, it's basically like `value.in(array)` if I understand it correctly. That reads nicely. But here (and in Catmandu) it seems more natural to me to say the container first after `in`, like 'if in array value' not 'if in value array'.

blackwinter:
Well, to a degree this applies to all the predicates, doesn't it? Everything has to be read with the first parameter as subject; you just have to make that translation on-the-fly. But I agree that this is a particularly egregious case. A preposition doesn't really lend itself to being used in prefix position.

Would you feel the same if `in` were renamed to `is_contained_in`? Maybe we could provide it as an alias?

PR #237, opened 2022-07-01
Add `once()` Fix bind.

blackwinter:
Allows to perform setup tasks that don't have to (and/or should not) be repeated for each record. In Limetrans we implemented it as a [custom Fix bind](https://github.com/hbz/limetrans/blob/55da0b6c1a706854637a755caa94e27aacfa8a6e/src/main/java/hbz/limetrans/function/Once.java) (cf. #124), but maybe it's useful generally.

Open question: Currently, the scope is per Fix invocation. Should it be per Fix _file_ instead? (Then we would change `shouldExecuteOnlyOncePerFixInstance()` to `shouldExecuteOnlyOncePerFixFile()`.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,6 +577,27 @@ do list(path: "<sourceField>", "var": "<variableName>")
 end
 ```
 
+#### `do once`
+
+Executes the statements only once (when the bind is first encountered), not repeatedly for each record.
+
+```perl
+do once()
+  ...
+end
+```
+
+In order to execute multiple blocks only once, tag them with unique identifiers:
+
+```perl
+do once("maps setup")
+  ...
+end
+do once("vars setup")
+  ...
+end
+```
+
 ### Conditionals
 
 Conditionals start with `if` in case of affirming the condition or `unless` rejecting the condition.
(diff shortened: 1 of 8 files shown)


fsteeg (approved):

fsteeg:
> Open question: Currently, the scope is per Fix invocation. Should it be per Fix file instead? (Then we would change shouldExecuteOnlyOncePerFixInstance() to shouldExecuteOnlyOncePerFixFile().)

What would the difference be in practice? In general, I'd say we go with what worked in the context that this came up.

blackwinter:
> What would the difference be in practice? In general, I'd say we go with what worked in the context that this came up.

We haven't used it more than "once" yet ;)

I guess, the difference would be when sharing Fix files? If you `include()` a component that has a `once()` block, any other `once()` blocks would be skipped. OTOH, if we would scope by file and you refactor multiple `include()`s into a single file, the previously executed `once()` blocks would now be skipped (except for one).

So best practice would probably be to always try to tag them with a unique identifier. Then it doesn't seem to matter which scope we choose. Although the current one is supposedly more predictable.

fsteeg:
> So best practice would probably be to always try to tag them with a unique identifier. Then it doesn't seem to matter which scope we choose. Although the current one is supposedly more predictable?

+1 for the current implementation.

TobiasNx:
+ 1 works for me.

PR #238, opened 2022-07-01
Add `debug_record()` Fix function.

blackwinter:
Allows to print the current record during transformation (even at different stages). In Limetrans we implemented it (in a simplified form) as a [custom Fix function](https://github.com/hbz/limetrans/blob/55da0b6c1a706854637a755caa94e27aacfa8a6e/src/main/java/hbz/limetrans/function/DebugRecord.java) (cf. #105), but maybe it's useful generally.

Tests are still missing as I'm unsure how to go about it. Any suggestions?

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,37 @@ Copies (or appends to) a field from an existing field.
 copy_field("<sourceField>", "<targetField>")
 ```
 
+##### `debug_record`
+
+Prints the current record either to standard output or to a file.
+
+Parameters:
+
+- `prefix` (optional): Prefix to print before the record. (Default: Empty string)
+
+Options:
+
+- `compression` (file output only): Compression mode. (Default: `auto`)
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -242,6 +242,37 @@ Copies (or appends to) a field from an existing field.
 copy_field("<sourceField>", "<targetField>")
 ```
 
+##### `debug_record`
+
+Prints the current record either to standard output or to a file.
```
TobiasNx, line 247:
```suggestion
Prints the current record at the position of the function either to standard output or to a file. It can help to debug the  fix-transformation.
```

TobiasNx:
This seems to be a nice feature.
Had a look at the documentation but did not try it yet.
I think I could try it in context of ALMA-FIX. I then could try it in the transformation

+1 for merging.

fsteeg:
> Actually, I'd be fine with having it [JSON] as the sole output format

+1

> Also, I think I'd like to change prefix to a format string just like destination

+1

blackwinter:
I've added the output of the internal representation back (be32875) as it's the only way to see the record without interference from the encoding process (e.g. in the context of investigating [this error](https://github.com/metafacture/metafacture-fix/pull/244#issuecomment-1185306366)).

PR #241, opened 2022-07-07
Remove path replacement workarounds, only set and append paths

fsteeg:
Append to path when adding values, set base path for nested values.

This seems to resolves the issue in hbz/lobid-resources#1354, see unit and integration tests, which are now passing, but tweaked (unit test for output structure, integration test for field order).

### DIFF
--- a/metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
+++ b/metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
@@ -33,7 +33,6 @@ import java.util.Arrays;
  * @author Fabian Steeg
  */
 @ExtendWith(MockitoExtension.class)
-@ExtendWith(MetafixToDo.Extension.class)
 public class MetafixLookupTest {
 
     private static final String CSV_MAP = "src/test/resources/org/metafacture/metafix/maps/test.csv";
@@ -789,7 +788,6 @@ public class MetafixLookupTest {
     }
 
     @Test
-    @MetafixToDo("See https://github.com/hbz/lobid-resources/pull/1354")
     public void shouldLookupInCopiedNestedArrays() {
         MetafixTestHelpers.assertFix(streamReceiver, Arrays.asList(
                 "put_map('rswk-indicator', s: 'SubjectHeading')",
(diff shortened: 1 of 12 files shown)


metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
```diff
@@ -819,22 +817,10 @@ public void shouldLookupInCopiedNestedArrays() {
...
-                o.get().literal("1", "SubjectHeading");
-                f.apply(2).endEntity();
-                o.get().startEntity("5");
-                o.get().startEntity("type[]");
-                o.get().literal("1", "SubjectHeading");
+                o.get().literal("2", "SubjectHeading");
+                o.get().literal("3", "SubjectHeading");
+                o.get().literal("4", "SubjectHeading");
+                o.get().literal("5", "SubjectHeading");
                 f.apply(5).endEntity();
```
blackwinter, line 824: > but tweaked (unit test for output structure, [...])

Hm, sorry about that. No idea how I came up with that expected output.

metafix/src/test/resources/org/metafacture/metafix/integration/lookup/fromXml/toJson/lookupInDeeplyNestedArrayOfObjects/test.fix
```diff
@@ -3,25 +3,24 @@ put_map("rswk-indicator", s: "SubjectHeading")
 set_array("subject[]")
 
 if exists("6890?")
-  set_array("subject[].$append.type[]", "ComplexSubject")
-  set_array("subject[].$last.label")
-  set_array("subject[].$last.componentList[]")
+  set_array("subject[].$append.componentList[]")
```
blackwinter: > but tweaked ([...], integration test for field order)

Sorry again. Looks like I accidentally sorted the expected output. Fixed in fe63bad.

blackwinter (approved):

metafix/src/test/resources/org/metafacture/metafix/integration/lookup/fromXml/toJson/lookupInDeeplyNestedArrayOfObjects/test.fix
blackwinter: > The original expected output was identical to the `fromJson` test, so maybe that's where that came from.

Yeah, in the `fromJson` test the input is sorted as well, while in the `fromXml` test the output is generated in the Fix.

PR #242, opened 2022-07-08
Add `from_json()`/`to_json()` Fix functions.

blackwinter:
Open questions:

1. `shouldConvertNullFromJson()` seems counterintuitive; but fixing it would require changing the behaviour of `Record.transform(String, BiConsumer)`. Also, we'd have to differentiate between value being `null` and `error_string` being `null`. One option might be to generate an empty string which could then get at least cleaned up by `vacuum()`.
2. `to_json()` error behaviour is untested; I have no idea how to conjure up an invalid object ;)
3. Should we somehow utilize JsonDecoder/JsonEncoder instead of interacting with Jackson directly? Would certainly allow for greater compatibility/configurability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,6 +497,18 @@ Flattens a nested array field.
 flatten("<sourceField>")
 ```
 
+##### `from_json`
+
+Replaces the string with its JSON deserialization.
+
+Options:
+
+- `error_string`: Error message as a placeholder if the JSON couldn't be parsed. (Default: `null`)
+
+```perl
+from_json("<sourceField>"[, error_string: "<errorValue>"])
+```
+
 ##### `index`
 
 Returns the index position of a substring in a field and replaces the field value with this number.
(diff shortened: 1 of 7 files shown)


fsteeg:
> So basically: How should JSON null behave? Produce string "null" (current behaviour) or delete itself from the record (because we don't accept null values)?

Alright, got it, thanks. I think the current behavior (produce string "null") is good.

fsteeg (approved):

TobiasNx:
Had a look at the function, the docu and the unit test.
Did not play around with it yet.

What would a use case for this look like? Perhaps a short description of a use case could be added in the docu as example. (not the code)

blackwinter:
> What would a use case for this look like?

Our actual use case for `from_json()` (in combination with `flatten()`, see #244) is `lookup()` in a multi-valued map where values are encoded as JSON (because maps can only have string values).

`to_json()` was more or less a by-product, but already proved itself useful for `print_record()` (see #238).

blackwinter:
> Then go for it! +1

You mean describing this rather complex use case in the documentation for `from_json()`? I'm not so sure.

blackwinter:
Okay, thanks!

PR #243, opened 2022-07-12
Add `put_macro()`/`call_macro()` Fix bind and function.

blackwinter:
Ports Metamorph [macros](https://github.com/metafacture/metafacture-core/wiki/Metamorph-User-Guide#defining-macros) to Metafix.

Open questions:

- Should the function be named `call_macro()` instead?
- Should calling an unknown macro throw an exception?

### DIFF
--- a/README.md
+++ b/README.md
@@ -234,6 +234,25 @@ array("foo")
 # {"name":"value"} => ["name", "value"]
 ```
 
+##### `call_macro`
+
+Calls a named macro, i.e. a list of statements that have been previously defined with the [`do put_macro`](#do-put_macro) bind.
+
+Parameters:
+
+- `name` (required): Unique name of the macro.
+
+Options:
+
+- All options are made available as "dynamic" local variables in the macro.
+
+```perl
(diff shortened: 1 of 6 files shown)


blackwinter:
There might be an issue with local variables not being available inside binds in a macro. I'll investigate.

fsteeg:
> Should the function be named `call_macro()` instead?

I think that's clearer. Also, how about `put_macro` for the bind, like `put_map` and `put_var`?

> Should calling an unknown macro throw an exception?

I think so, otherwise a typo in the macro name would cause it to be silently skipped, right?

blackwinter:
Thanks for your feedback. I've applied it in 1d06fef.

fsteeg (approved):

blackwinter:
> I would add in the `README.md` how to define and call the variables in the macro.

They are just passed as options and used like any other (global) variables. What would you suggest?

TobiasNx (approved):
One suggestion for the README.md otherwise +1 nice!

TobiasNx:
+1

PR #244, opened 2022-07-12
Add `flatten()` Fix function.

### LINKED ISSUE #122: Add `flatten`-function
Flatten is not added as function.
It can be helpful if we have an array of keywords with some keywords connected and some tokenized.

in:
```
{
	"keywords": [
		"dog",
		"cat",
		"horse,bird,dragon"
	]
}
```

If we would split the third keyword we would (which we cannot at the moment) we would get an array in an array. Then `flatten` would be nice.

result (does not work yet because of the asterisk-problem with spli-field):
```
{
	"keywords": [
		"dog",
		"cat",
		["horse" , "bird" , "dragon"]
	]
}
```
should:
```
{
	"keywords": [
		"dog",
		"cat",
		"horse",
		"bird",
		"dragon"
	]
}
```

See test-scenario:
https://github.com/TobiasNx/fix-FunctionalReview-Testing/tree/master/data/testing/flattening

blackwinter:
Implements Catmandu's [`flatten()`](https://metacpan.org/pod/Catmandu::Fix::flatten) function.

Resolves #122.

### DIFF
--- a/README.md
+++ b/README.md
@@ -457,6 +457,14 @@ Only keeps field values that match the regular expression pattern.
 filter("<sourceField>", "<regexp>")
 ```
 
+##### `flatten`
+
+Flattens a nested array field.
+
+```perl
+flatten("<sourceField>")
+```
+
 ##### `index`
 
 Returns the index position of a substring in a field and replaces the field value with this number.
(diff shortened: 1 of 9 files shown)


fsteeg (approved):

blackwinter:
> Don't know if this is connected to flatten or a general issue.

No, it's a general issue. It even worked with `nothing()` as the Fix.

Metafix wasn't able to emit nested arrays properly (see #248 for a fix).

PR #245, opened 2022-07-12
Pass `include()` options as variables.

blackwinter:
Missing piece for #12.

_Base branch will be changed automatically once #243 is merged._

PR #246, opened 2022-07-13
Add `str_contain()`/`str_equal()`/`str_match()` Fix conditionals.

blackwinter:
Complements path/string (`any_*()`) and path/path (`in()`) conditionals with string/string variants.

Particular useful in combination with local variables in `macro()` (#243) or `include()` (#245).

_Base branch will be changed automatically once #243 is merged._

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/FixConditional.java
+++ b/metafix/src/main/java/org/metafacture/metafix/FixConditional.java
@@ -41,6 +41,12 @@ public enum FixConditional implements FixPredicate {
             return !any_contain.test(metafix, record, params, options);
         }
     },
+    str_contain {
+        @Override
+        public boolean test(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
+            return testConditional(params, CONTAINS);
+        }
+    },
 
     all_equal {
         @Override
@@ -58,6 +64,12 @@ public enum FixConditional implements FixPredicate {
         @Override
         public boolean test(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
             return !any_equal.test(metafix, record, params, options);
+        }
+    },
+    str_equal {
+        @Override
+        public boolean test(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
+            return testConditional(params, EQUALS);
         }
     },
 
@@ -112,6 +124,12 @@ public enum FixConditional implements FixPredicate {
         @Override
         public boolean test(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
             return !any_match.test(metafix, record, params, options);
+        }
+    },
+    str_match {
+        @Override
+        public boolean test(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
+            return testConditional(params, MATCHES);
         }
     }
 
--- a/metafix/src/main/java/org/metafacture/metafix/api/FixPredicate.java
+++ b/metafix/src/main/java/org/metafacture/metafix/api/FixPredicate.java
@@ -49,4 +49,8 @@ public interface FixPredicate {
         ));
     }
 
+    default boolean testConditional(final List<String> params, final BiPredicate<String, String> conditional) {
+        return conditional.test(params.get(0), params.get(1));
+    }
+
 }
--- a/metafix/src/test/java/org/metafacture/metafix/MetafixIfTest.java
+++ b/metafix/src/test/java/org/metafacture/metafix/MetafixIfTest.java
@@ -1430,4 +1430,179 @@ public class MetafixIfTest {
         );
     }
 
+    @Test
+    public void shouldContainString() {
+        MetafixTestHelpers.assertFix(streamReceiver, Arrays.asList(
+                "if str_contain('name', 'a$[var]')",
+                "  add_field('type', 'Organization')",
+                "end"
+            ),
+            ImmutableMap.of("var", "me"),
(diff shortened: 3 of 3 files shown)


fsteeg:
Took a bit for me to understand that we're actually comparing literal strings here. I'm wondering if there might be a way to make this more obvious. One thing is probably the tests, which use `name` as the literal string and as a field name. Maybe different names could also help, something like `literal_*`?

Or we could use the `~` prefix from the `paste` method to mark literals in the existing methods? The problem here is that the `any_`, `all_` and `none_` prefixes make no sense with the literal values. Then again I often found it cumbersome having to think about any/all in the common case of using it on single-value fields.

So I guess my suggestion would be to create `contain`, `equal` and `match` methods which alias their `any_` versions and support literal strings prefixed with `~`.

blackwinter:
> Took a bit for me to understand that we're actually comparing literal strings here. I'm wondering if there might be a way to make this more obvious.

Hm, it's literally in the description: "string/string variants". But I guess that wasn't obvious enough ;)

> One thing is probably the tests, which use `name` as the literal string and as a field name.

That's intentional, to demonstrate that the outcome is independent of the field's value.

Let me think about your other suggestions.

blackwinter:
> Or we could use the `~` prefix from the `paste` method to mark literals in the existing methods?

I have to admit, I'm totally not a fan of markers in general (this includes our ever more pervasive entity markers; _especially_ when there's no way to escape them). So I guess I'm against any proliferation of this "bad" habit (well, except where we don't really have a better alternative? E.g. metafacture/metafacture-core#458).

> The problem here is that the `any_`, `all_` and `none_` prefixes make no sense with the literal values. Then again I often found it cumbersome having to think about any/all in the common case of using it on single-value fields.

I, for one, try not to think of fields as single-valued or multi-valued. Every field (more or less) has the potential to be either, it all depends on the data.

> Maybe different names could also help, something like `literal_*`?

That we can certainly do. I've contemplated this option for a while before settling on `str_*`, just because it seemed to fit a little better with `any_*` (and also because I had the impression that the term "literal" isn't as established in the Fix context as it is in Metafacture).

To be honest, though, my initial implementation was geared towards the variable: `var_*` with the first parameter being a variable name, not a string literal with potential variable interpolations. Do you think this would be a better approach?

----

BTW: Where is the documentation for the new conditionals? I could have sworn I added them to the README (should have also helped with the initial confusion). I'll do it as soon as we've decided on a path forward.

fsteeg:
> I have to admit, I'm totally not a fan of markers in general [...] So I guess I'm against any proliferation of this "bad" habit

I see your point, but in this case I'd say that we already have a convention for this exact case of differentiating a literal value from a field name, so it would improve consistency using it. If designing from scratch we'd probably have introduced a syntactic way to differentiate literal values and field names, but this is the approach Catmandu takes. I also think the `~` prefix makes it very clear that the difference is actually in the first param only:

`match('name', '$[var]')` vs. `match('~name', '$[var]')`

Compared to having different methods which actually only differ in the way the first param is handled:

`any_match('name', '$[var]')` vs. `str_match('name', '$[var]')`

But yeah, this involves the introduction of the aliases which is another thing to discuss, and you're not in favor of that too, so I really don't insist on this approach here.

> [...] `var_*` with the first parameter being a variable name, not a string literal with potential variable interpolations.

Ah, so `var_match('var', 'name')` instead of `str_match('name', '$[var]')`?

I like it because this actually *is* a completely different method (not only in the way the first param is handled), and the literal value is in its usual, second position, like in the other conditionals.

blackwinter:
> I see your point, but in this case I'd say that we already have a convention for this exact case of differentiating a literal value from a field name, so it would improve consistency using it.

You say precedence, I say proliferation ;) Yes, we have a convention (based on a _single_ occurrence, mind you), and yes, we could apply it here. I also don't want to dismiss your idea of combining `any_*` and `str_*` into prefix-less conditionals. But I'll need more time to think about it.

> I also think the `~` prefix makes it very clear that the difference is actually in the first param only:
> [...]
> Compared to having different methods which actually only differ in the way the first param is handled:

That's what I actually like about this approach :) There's a symmetry in what the tests refer to:

- `any_*()`: path/string
- `in()`: path/path
- `str_*()` string/string

> Ah, so `var_match('var', 'name')` instead of `str_match('name', '$[var]')`?

Not quite. The equivalent of `str_match('name', '$[var]')` would be `var_match('name_var', '$[var]')` with another variable `name_var: "name"`. A better illustration would probably be `str_match('$[var]', 'name')` <=> `var_match('var', 'name')`.

> I like it because this actually _is_ a completely different method (not only in the way the first param is handled), and the literal value is in its usual, second position, like in the other conditionals.

It's less flexible and _feels_ kinda arbitrary. But I think I like it better than the `~` marker ;)

Let's we give it a little more time. And also @TobiasNx a chance to weigh in.

TobiasNx:
It also took me a while to figure out what this does, even if it seems obvious.
Had a look at the unit tests and I think I understand it now.

At the moment I do not know of any scenario I would use it (yet!) since the scenarios are very sophisticated and are in combination with variables and makros.

I think I favour `str_` a little bit more than `var_` but I also would dislike `~`.
I do not want to be the one to decide.

fsteeg (approved):
Let's go with the current `str_` implementation then :+1:

TobiasNx (approved):
+1

blackwinter:
Thank you both for the discussion!

blackwinter:
And now I forgot the documentation :( I'll add it separately.

PR #247, opened 2022-07-14
Add `is_<type>` Fix conditionals.

blackwinter:
Implements a bunch of type-related conditionals from Catmandu.

- Adds `is_hash()` as alias for `is_object()`.
- Adds `is_empty()` as new conditional.

Differs from Catmandu in that all scalars in Metafix are strings, we can't actually distinguish between boolean values and boolean strings or numeric values and numeric strings (that's also why `is_true()`/`is_false()` are hard-wired to `strict: false`).


TobiasNx:
This integration test seems not to work: https://github.com/metafacture/metafacture-fix/pull/247/commits/29065de54ae60b7ed7fd45f26a5d305f114a7362

TobiasNx:
```
#### `is_string`

Executes the functions if/unless the field value is a string (and not a number).
```

saw this what about numbers that are strings? i think this is difficult to differentiate.

I would have not expected that the conditional is false,  would have expected this do be true
https://github.com/metafacture/metafacture-fix/blob/77f6a97b4322b32e2327a7515cdb2eac66fc24af/metafix/src/test/resources/org/metafacture/metafix/integration/conditional/fromJson/toJson/ifIsStringsButNumbers

blackwinter:
That's just not how Catmandu behaves. Do you suggest breaking compatibility here?

TobiasNx:
> That's just not how Catmandu behaves. Do you suggest breaking compatibility here?

perhaps an option then? 

```
if is_string("elementA", numberStrings:"true")
  add_field("elementB", "Hello")
end
```

fsteeg:
> perhaps an option then?

I think the natural way to express that would be if we had boolean operators (https://github.com/metafacture/metafacture-fix/issues/165):

`if is_string(elementA) || is_number(elementB)`

In the meantime, a workaround could be to call a macro from both `if` and `elsif`:

```
if is_string(elementA)
  call_macro(macroA)
elsif is_number(elementA)
  call_macro(macroA)
end
```

blackwinter:
Yes, boolean operators are at the top of my wish list now :)

Would you two agree to accept the implementation as is and maybe change it later depending on the outcome of LibreCat/Catmandu#393?

TobiasNx:
> > Would you two agree to accept the implementation as is and maybe change it later depending on the outcome of [LibreCat/Catmandu#393](https://github.com/LibreCat/Catmandu/issues/393)?
> 
> +1



> > Would you two agree to accept the implementation as is and maybe change it later depending on the outcome of [LibreCat/Catmandu#393](https://github.com/LibreCat/Catmandu/issues/393)?
> 
> +1

+1 also all seem to work

fsteeg (approved):

blackwinter:
> saw this what about numbers that are strings? i think this is difficult to differentiate.

Maybe we missed this earlier, but isn't the issue that Metafix doesn't have a numeric data type? Internally, all scalar values are strings.

PR #248, opened 2022-07-15
Fix emitting nested arrays.

blackwinter:
Nested arrays would [always be flattened](https://github.com/metafacture/metafacture-fix/pull/244#issuecomment-1185306366).


fsteeg (approved):

TobiasNx:
Should I test this beside the flatten test since this was a global issue?

blackwinter:
Up to you. This has the potential to break existing uses. But it's a bug fix after all. So if you have anything you want to ensure keeps working, then you should test that.

TobiasNx:
I tested it with  my ALMA-Transformation had no breaks or changes. So seems good :)

TobiasNx (approved):

PR #254, opened 2022-08-30
Add include tests for testing

TobiasNx:


blackwinter:
What's the status of this pull request?

TobiasNx:
@blackwinter  I assume this integration test should test if include works and compare it to a version with a single fix. I am not sure if this is still necessary. What do you think. I would close it if not needed.

blackwinter:
Seeing as we already have unit tests for this functionality, I don't think it's strictly necessary. But you opened this pull request as a draft and never put it up for review. That's why I asked what its status was. So I would suggest to either mark it as "Ready for review" (and assign reviewers) or close it. Up to you.

Issue #255, opened 2022-09-05
Race conditions when using Threads


dr0i:
Came up in https://github.com/hbz/lobid-resources/issues/1409#issuecomment-1235382923.
This one is not reproducible and seems to be the result of Metafix not being thread safe:

> WARN [Thread-4] (Metafix.java:421) - Error while executing Fix expression (at file:/home/sol/git/lobid-resources-alma-fix/src/main/resources/alma/fix/contribution.fix, line 184): do list(path:"110[012] ", "var":"$i") unless exists("$i.M") unless exists("$i.a.1") unless exists("$i.4") if any_match("type[]","ArchivedWebPage|Miscellaneous|Bibliography|Statistics|Legislation|PublishedScore|Game|Image|Map|Standard") add_field("$i.4","cre") elsif any_match("type[]","Periodical|Collection|Series|Newspaper|Journal|PublicationIssue|EditedVolume|Proceedings|Festschrift|ReferenceSource") add_field("$i.4","edt") elsif any_match("type[]","Book|MultiVolumeBook|Article|Thesis|OfficialPublication|Schoolbook|Biography|Report") add_field("$i.4","aut") else add_field("$i.4","cre") end end do list(path: "$i.4", "var":"$j") set_hash("contribution[].$append.agent") do list(path:"$i.0","var":"$k") if all_match("$k", "^\\(DE-588\\).*$") paste("contribution[].$last.agent.gndIdentifier","$k") paste("contribution[].$last.agent.id","$k") end end paste("contribution[].$last.agent.label", "$i.a", "$i.b", join_char: ". ") set_array("contribution[].$last.agent.type[]","CorporateBody") copy_field("$j","contribution[].$last.role.id") end end end end
java.util.ConcurrentModificationException
        at java.util.HashMap$HashIterator.nextNode(HashMap.java:1469)
        at java.util.HashMap$KeyIterator.next(HashMap.java:1493)
        at org.metafacture.commons.tries.WildcardTrie.matches(WildcardTrie.java:128)
        at org.metafacture.commons.tries.WildcardTrie.get(WildcardTrie.java:123)
        at org.metafacture.commons.tries.SimpleRegexTrie.get(SimpleRegexTrie.java:75)
        at org.metafacture.metafix.Value$Hash.findFields(Value.java:806)
        at org.metafacture.metafix.Value$Hash.findFields(Value.java:778)
        at org.metafacture.metafix.Value$Hash.containsField(Value.java:522)
[...]

PR #257, opened 2022-09-12
255 make thread safe

dr0i:
Helps to prevent race conditions.

See https://github.com/metafacture/metafacture-fix/issues/255 .
See also https://github.com/metafacture/metafacture-core/pull/465 for adaptions needed in dependency `metafacture-core`.

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -140,7 +140,10 @@
         </module>
         <module name="SimplifyBooleanExpression"/>
         <module name="SimplifyBooleanReturn"/>
-        <module name="StaticVariableName"/>
+        <module name="StaticVariableName">
+            <!-- disallow static, non-final variables -->
+            <property name="format" value="^$"/>
+        </module>
         <module name="StringLiteralEquality"/>
         <module name="SuperClone"/>
         <module name="SuperFinalize"/>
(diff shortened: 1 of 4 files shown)


blackwinter:
General question (also applies to metafacture/metafacture-core#465): Can you expand a little on your rationale? How did you arrive at these specific synchronization points? Some of them don't make sense to me at first glance.

Also: Can you document in #255 the cause for each error you got?

dr0i:
Sure: at first I followed the exceptions described in #255. Further Exceptions appear nonetheless, eliminating them with further `synchronized`. As theses Exceptions appear after a longer time of ETL of the basedump, I put in multiple `synchronized` to go on quicker. At one point I tried to boil down where exactly a `synchronized` would make sense, but because of the needed time to check these (unpredictable) race conditions that became to time intensive. I experienced no problem with having _too much_ of `synchronized` rather than _to little_ I decided to put them everywhere where they might make sense.
I appreciate your review and there might well be a dozen of superflous `synchronized`, you may remove them at your will.

dr0i:
This breaks down performance significantly. More brain has to put into that, reducing the needed synchronized to a minimum. Or maybe dump the idea of fixing this at all - loss of documents seems not to be that big.

blackwinter:
If we can find the root cause(s) we might be able to put in a targeted fix. But I agree wholeheartedly, sprinkling `synchronized` all over the place might not be the best way to go about it ;)

dr0i:
Closed in favor of https://github.com/metafacture/metafacture-fix/pull/263.

PR #259, opened 2022-09-16
Fix macro performance regression.

### LINKED ISSUE #258: Performance problem: exponential running time
Since a month or so we experience an exponential running time, regardless using threads or not. Here a table from the logs:

|timestamp| time compared to predecessor| every 100k docs|
| ---      | ---       | ---- |
|2022-09-12 17:30:44| +30m | 100000|
|2022-09-12 18:31:59| +60m| 200000|
|2022-09-12 20:03:10| +90m| 300000|
|2022-09-12 22:05:05| +120m|  400000|
|2022-09-13 00:39:17| +150m| 500000|
|2022-09-13 03:45:12| +180m| 600000|
|2022-09-13 07:16:26| +210m| 700000|
|2022-09-13 11:20:18| +240m| 800000|

=> every 100kth docs the time needed increased by 30m in comparison to the 100kth predecessor.
One speculation regarding the cause:
some data structures are `static` ( particularly `Trie` and `HashMap` as cache) which might not being removed by the Garbage Collector (GC) and thus grow with time. Especially operations on the `Trie` become slower with its growth. Solution would be to guarantee proper GC of `Trie`.

See useful hints for analyzing at https://github.com/metafacture/metafacture-fix/issues/207.

blackwinter:
Every time a macro is (re-)declared (e.g., for each record), its "static" local variables (options) are added to an internal list. So this list grows with each record and merging it (`RecordTransformer.getVars()`) takes longer and longer.

Instead, keep track of each type of variables (global, static, dynamic) separately and replace already existing variables of the same type.

Resolves #258.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/FixBind.java
+++ b/metafix/src/main/java/org/metafacture/metafix/FixBind.java
@@ -74,7 +74,7 @@ public enum FixBind implements FixContext {
     put_macro {
         @Override
         public void execute(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options, final RecordTransformer recordTransformer) {
-            recordTransformer.addVars(options);
+            recordTransformer.setVars(options);
             metafix.putMacro(params.get(0), recordTransformer);
         }
     }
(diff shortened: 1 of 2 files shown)


dr0i (approved):
That came fast! Thx for fixing this!

PR #260, opened 2022-09-16
Include caller location and text in macro execution exception.

blackwinter:
When a statement inside a macro fails, it's hard to find the actual call site. Provide the caller location for easier debugging.

Open question:

- Should the introductory phrase (`Error while ...`) be included for every item in the stack? Or should we build the combined message differently?

Related to #123 and #243.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/Metafix.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Metafix.java
@@ -171,7 +171,13 @@ public class Metafix implements StreamPipe<StreamReceiver>, Maps {
     }
 
     public RecordTransformer getMacro(final String name) {
-        return macros.get(name);
+        final RecordTransformer macro = macros.get(name);
+
+        if (macro != null) {
+            macro.setParentExceptionMessage(recordTransformer);
+        }
+
+        return macro;
     }
 
     public List<Expression> getExpressions() {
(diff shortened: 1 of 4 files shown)


metafix/src/main/java/org/metafacture/metafix/Metafix.java
```diff
@@ -171,7 +171,13 @@ public void putMacro(final String name, final RecordTransformer macro) {
     }
 
     public RecordTransformer getMacro(final String name) {
-        return macros.get(name);
+        final RecordTransformer macro = macros.get(name);
+
+        if (macro != null) {
+            macro.setParentExceptionMessage(recordTransformer);
```
fsteeg, line 177: Minor: I find the `set...` naming where the argument is not the thing that is set slightly confusing. Maybe `setParentExceptionMessageFrom`? Or instead of a helper method, the single-line implementation could go right here.

fsteeg (approved):
> Should the introductory phrase (Error while ...) be included for every item in the stack?

I think that's fine.

metafix/src/main/java/org/metafacture/metafix/Metafix.java
blackwinter: I see what you mean. Changed in 36cd3da.

> Or instead of a helper method, the single-line implementation could go right here.

But then we would have to expose `currentMessageSupplier`, which I'd rather not.

PR #269, opened 2022-11-16
Add `lookup()` option to print unknown values.

### LINKED ISSUE #212: Add lookup option to save not matching values in extra file
Could `lookup` be enhanced to export not matching values into a separate file?
This is good to track changing vocabularies.

`lookup("title","dict.csv", sep_char:",", exportNotMatching:"list.txt.")`

blackwinter:
Resolves #212.


blackwinter:
Oh, you were too quick ;) I forgot to mention that there's a subtle behaviour change here: If a key is associated with a `null` value, it's logged as unknown and the default value is returned (if given, otherwise the key is returned); previously, the key would have been returned even with a default value present. This only applies to Java maps, of course, since Fix doesn't have `null` values.

blackwinter:
Heads up: I _think_ this won't work across multiple records (unless a separate file is written for each record, using counter and/or record ID). I have yet to verify, but am pretty sure ;) (see also metafacture/metafacture-core#475)

README.md
```diff
@@ -543,10 +543,39 @@ join_field("<sourceField>", "<separator>")
...
 Looks up matching values in a map and replaces the field value with this match. External files as well as internal maps can be used.
 
+Parameters:
+
+- `path` (required): Field path to look up.
+- `map` (optional): Name or path of the map in which to look up values.
+
+Options:
+
+- `__default`: Default value to use for unknown values. (Default: Old value)
```
TobiasNx, line 553: is this a double `_` ?

blackwinter: Yes; see [Maps.DEFAULT_MAP_KEY](https://javadoc.io/doc/org.metafacture/metamorph-api/5.4.0/constant-values.html#org.metafacture.metamorph.api.Maps.DEFAULT_MAP_KEY).

blackwinter:
Fixed printing unknown values across multiple records in 356a5e7.

(Also rebased on top of master, i.e. 5d756eb)

TobiasNx:
The append option is a little bit unclear. If `append:"false"` assumed that the file is overwritten per metafacture run not per record. But it seems that the file is overwritten per record.

blackwinter:
> If `append:"false"` assumed that the file is overwritten per metafacture run not per record.

No, with append mode disabled, the file is overwritten for every `lookup()` call. Otherwise, it gets appended to instead of overwritten.

> But it seems that the file is overwritten per record.

Exactly, that's the expected behaviour.

TobiasNx:
> > If `append:"false"` assumed that the file is overwritten per metafacture run not per record.

anyway to make this a feature?
Use case: In OERSI every day all records are transformed. Therefore `append` would generate a lot of duplicates because of the daily transformation of each record.

blackwinter:
> anyway to make this a feature?

What do you have in mind? A timestamp in the file name?

> Therefore `append` would generate a lot of duplicates because of the daily transformation of each record.

But you are going to add unknown values over time, aren't you? And then you'd delete the file.

Otherwise, I would suggest to just rotate the file.

blackwinter:
> I thought to overwrite just once, the first time the lookup is run.

We can try if that's the requirement.

@fsteeg: Want to weigh in?

TobiasNx:
> > I thought to overwrite just once, the first time the lookup is run.
> 
> We can try if that's the requirement.
> 
> @fsteeg: Want to weigh in?

If this is to much hastle. Let's go with this as it is for now, and evaluate later if this is a needed feature! 


One further thing. I tried to set id to name but the printed output has not the value of name in it:
https://github.com/metafacture/metafacture-fix/pull/270

blackwinter:
> One further thing. I tried to set id to name but the printed output has not the value of name in it: #270

1. The record ID is not included in the output, it's only meant for inclusion in the file name (`destination`).
2. It's actually not implemented as documented right now. Going to fix.

blackwinter:
Fixed `destination` and also added `prefix` option (same as `print_record()`'s `prefix` parameter).

blackwinter:
> Let's go with this as it is for now, and evaluate later if this is a needed feature!

That's fine with me, but we should keep in mind that such a change _might_ not be possible in a backwards-compatible manner.

PR #274, opened 2022-12-02
Update README.md

TobiasNx:
documented that `uniq``also works with array of objects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -687,7 +687,7 @@ trim("<sourceField>")
 
 ##### `uniq`
 
-Deletes duplicate values in an array.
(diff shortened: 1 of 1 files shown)


blackwinter (changes requested):
In our documentation, so far, "values" are generally strings, arrays or objects (hashes); we don't (usually) differentiate. I'd be -0.5 w.r.t. this change, unless you want to update all uses of "value" and specify whether it refers to strings, arrays and/or hashes/objects.

As a matter of fact, `uniq()` works with all types of values (in an array).

Issue #231, opened 2022-06-22
Add `isbn`-function as in morph


TobiasNx:
Morph has a specific [morph function for validating and transforming ISBNs](https://github.com/metafacture/metafacture-core/blob/master/metamorph/src/main/java/org/metafacture/metamorph/functions/ISBN.java).

This is used in the ALMA Transformation of lobid. We need a similar function in FIX.

blackwinter:
Functional review: @TobiasNx
Code review: @fsteeg

TobiasNx:
+1 great thanks

PR #285, opened 2023-02-17
Add uri_encode function (#273)

### LINKED ISSUE #273: Add function to percent-encode strings
Catmandu has:
```

# '3%A9' => 'café'
uri_decode(place)
# 'café' => '3%A9'
uri_encode(place)
```

MF Morph has `urlencode`(https://github.com/metafacture/metafacture-core/blob/master/metamorph/src/main/java/org/metafacture/metamorph/functions/URLEncode.java):

```
<data source="suchwert">
  <urlencode />
  <compose prefix="https://swissbib.ch/Search/Results?lookfor="/>
</data>
```

We need this for formatting/encoding ids in lobid:

```
"hasItem" : [ {
    "label" : "lobid Bestandsressource",
    "type" : [ "Item", "PhysicalObject" ],
    "callNumber" : "61 TYD 16(3)",
    "currentLocation" : "JUE / JUE_Frei",
    "heldBy" : {
      "id" : "http://lobid.org/organisations/DE-A96#!",
      "isil" : "DE-A96",
      "label" : "lobid Organisation"
    },
    "id" : "http://lobid.org/items/990223521400206441:DE-A96:61 TYD 16(3)#!"
  }, {
    "label" : "lobid Bestandsressource",
    "type" : [ "Item", "PhysicalObject" ],
    "callNumber" : "61 TYD 16(3)",
    "currentLocation" : "JUE / JUE_Frei",
    "heldBy" : {
      "id" : "http://lobid.org/organisations/DE-A96#!",
      "isil" : "DE-A96",
      "label" : "lobid Organisation"
    },
    "id" : "http://lobid.org/items/990223521400206441:DE-A96:61 TYD 16(3)#!"
  },
```

dr0i:
Resolves #273.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/FixMethod.java
+++ b/metafix/src/main/java/org/metafacture/metafix/FixMethod.java
@@ -23,6 +23,8 @@ import org.metafacture.metamorph.functions.ISBN;
 import org.metafacture.metamorph.functions.Timestamp;
 import org.metafacture.metamorph.maps.FileMap;
 
+import com.google.common.net.PercentEscaper;
+
 import java.io.File;
 import java.io.IOException;
 import java.util.Arrays;
@@ -655,8 +657,15 @@ public enum FixMethod implements FixFunction { // checkstyle-disable-line ClassD
         public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
             record.transform(params.get(0), s -> s.toUpperCase());
         }
+    },
+    uri_encode {
+        @Override
+        public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
+            record.transform(params.get(0), PERCENT_ESCAPER::escape);
+        }
     };
 
+    private static final PercentEscaper PERCENT_ESCAPER = new PercentEscaper("", false);
     private static final Pattern NAMED_GROUP_PATTERN = Pattern.compile("\\(\\?<(.+?)>");
 
     private static final String FILEMAP_SEPARATOR_OPTION = "sep_char";
--- a/metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
+++ b/metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
@@ -3961,4 +3961,40 @@ public class MetafixMethodTest {
         );
     }
 
+    @Test
+    public void shouldUriEncodeString() {
+        MetafixTestHelpers.assertFix(streamReceiver, Arrays.asList(
+                "uri_encode('title')"
+            ),
+            i -> {
+                i.startRecord("1");
+                i.literal("title", "café");
+                i.endRecord();
+            },
+            o -> {
+                o.get().startRecord("1");
+                o.get().literal("title", "caf%C3%A9");
+                o.get().endRecord();
+            }
+        );
+    }
+
+    @Test
+    public void shouldUriEncodePathSegment() {
+        MetafixTestHelpers.assertFix(streamReceiver, Arrays.asList(
+                "uri_encode('id')"
+            ),
+            i -> {
+                i.startRecord("1");
+                i.literal("id", "slash/990223521400206441:DE-A96:61 TYD 16(3)#!");
+                i.endRecord();
+            },
+            o -> {
+                o.get().startRecord("1");
+                o.get().literal("id", "slash%2F990223521400206441%3ADE%2DA96%3A61%20TYD%2016%283%29%23%21");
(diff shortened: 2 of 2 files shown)


metafix/src/main/java/org/metafacture/metafix/FixMethod.java
```diff
@@ -655,8 +657,15 @@ public void apply(final Metafix metafix, final Record record, final List<String>
...
         }
+    },
+    uri_encode {
+        @Override
+        public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
+            record.transform(params.get(0), PERCENT_ESCAPER::escape);
+        }
     };
 
+    private static final PercentEscaper PERCENT_ESCAPER = new PercentEscaper("", false);
```
fsteeg, line 668: From what I understood, the conclusion from #273 was to basically use the [metamorph implementation](https://github.com/metafacture/metafacture-core/blob/master/metamorph/src/main/java/org/metafacture/metamorph/functions/URLEncode.java), which uses `java.net.URLEncoder`. If there is no specific reason for using `PercentEscaper` here, I think using the standard Java class would be preferable.

fsteeg (commented):
Added one inline comment on the encoding implementation.

We'd also have to add documentation on the new function in the README.

blackwinter:
In general, I would prefer to use the Metamorph implementation directly (see e.g. #230 and #231). If it doesn't provide a suitable interface it should be added there and then reused for both. Eventually, `org.metafacture.metamorph.functions` might be generalized and moved to `org.metafacture.commons.functions` (or something like that).

metafix/src/main/java/org/metafacture/metafix/FixMethod.java
dr0i: In lobid-resources I added a [UrlEscaper}(https://github.com/hbz/lobid-resources/blob/master/src/main/java/org/lobid/resources/UrlEscaper.java) which also complies to RFC  3986, using an older `com.google.gdata.util.common.base.PercentEscaper`. It hasn't make it to metafacture (yet).

dr0i:
> In general, I would prefer to use the Metamorph implementation directly

good point. Agreed. Awaiting ok for using percent encoder and then migrate function to metafcature-core.
> We'd also have to add documentation on the new function in the README.

Right - forgot about that. Would do this also im metfacture-core. Will note the RFC 3986 compliance there.

fsteeg:
One thing leading to confusion here (at least for me) is that we actually need something to encode a string to be used in a URI (in particular as a path segment), not to encode a full URI. So for `http://perl.com/foo bar` we don't want `http://perl.com/foo%20bar` (which would be a valid URI), but we want `http%3A%2F%2Fperl.com%2Ffoo%20bar` (which can be used as a path segment in a URI, but is not itself a valid URI).

Now if I understand it correctly, both [metamorph.functions.URLEncode](https://github.com/metafacture/metafacture-core/blob/master/metamorph/src/main/java/org/metafacture/metamorph/functions/URLEncode.java) and [Catmandu::Fix::uri_encode](https://metacpan.org/pod/Catmandu::Fix::uri_encode) basically do that, but Java's `URLEncoder` (which is currently used in `metamorph.functions.URLEncode`) uses `+` instead of `%20` for blanks, which is a problem for using it in lobid-resources.

To both avoid breaking the current `metamorph.functions.URLEncode` behavior and reusing as much implementation as possible, this is what @dr0i and I just discussed: we could base `metamorph.functions.URLEncode` on `PercentEscaper`, with an option to pass the `safeChars` config. In the default used in `metamorph`, we use [SAFECHARS_URLENCODER](https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.util.escape.PercentEscaper#com_google_api_client_util_escape_PercentEscaper_SAFECHARS_URLENCODER), which "mimics the behavior of java.net.URLEncoder" (thus no break), and in `metafix` we use e.g. [SAFE_PLUS_RESERVED_CHARS_URLENCODER](https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.util.escape.PercentEscaper#com_google_api_client_util_escape_PercentEscaper_SAFE_PLUS_RESERVED_CHARS_URLENCODER) (to get what we need in lobid-resources and be consistent with `Catmandu::Fix::uri_encode`).

blackwinter (changes requested):
Agreed. For increased flexibility the `safeChars` option should be exposed to Fix and Morph.

dr0i:
This PR is substituted by https://github.com/metafacture/metafacture-fix/pull/314.
Closing.

PR #297, opened 2023-04-27
Update vsc extension to new fix

### LINKED ISSUE #4: Fix comment syntax in vs code/codium extension (VSC)
Like `# This is a comment`, should be colored in vs code/codium extension.

### LINKED ISSUE #129: VS Code (VSC) - Wrong block outcommenting
When outcomment a block with key short cut  it results in `\* add_field("coll[].*.c", "test") */`. This is wrong

Also we decided for compatibility reason to catmandu to use `#` at the beginning of each line and not just  `/* */`

### LINKED ISSUE #291: Update VSC Extension
Due to changes in Metafacture Fix the VSC Extension should be updated by using the latest Fix version and adapting the syntax highlighting.
[This](https://github.com/hbz/lobid-resources/tree/master/src/main/resources/alma/fix) can be used for functional review.
The pictures at https://metafacture.org/ide-extensions/index.html should be exchanged.

functional review: @TobiasNx 
code review: @fsteeg

katauber:
Update vsc extension to match the current fix implementation.
- update highlighting of fix methods, fix conditionals and fix binds. Resolves #291 
- update line comments (#), remove wrong block comments. Resolves #4 and resolves #129

### DIFF
--- a/metafix-vsc/fix.configuration.json
+++ b/metafix-vsc/fix.configuration.json
@@ -1,6 +1,6 @@
 {
     "comments": {
-        "blockComment": [ "\\*", "*/" ]
+        "lineComment": "#"
     },
     "brackets": [
         ["{", "}"],
--- a/metafix-vsc/fix.tmLanguage
+++ b/metafix-vsc/fix.tmLanguage
@@ -32,7 +32,7 @@
 			<key>match</key>
 			<string>\b(all_contain|any_contain|none_contain|str_contain|all_equal|any_equal|none_equal|str_equal|exists|in|is_contained_in|is_array|is_empty|is_false|is_hash|is_number|is_object|is_sring|is_true|all_match|any_match|none_match|str_match)\b</string>
 			<key>comment</key>
-			<string>The names of the conditionals are taken from https://github.com/metafacture/metafacture-fix/blob/master/metafix/src/main/java/org/metafacture/metafix/FixMethod.java</string>
+			<string>The names of the conditionals are taken from https://github.com/metafacture/metafacture-fix/blob/master/metafix/src/main/java/org/metafacture/metafix/FixConditional.java</string>
 		</dict>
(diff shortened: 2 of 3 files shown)


fsteeg (approved):
Looks good. Just a bit confused about some names here (do these hook into some VSC logic?):

- `entity.name.function.fix` for enum values of `FixMethod`
- `keyword.operator.fix` for enum values of `FixConditional`
- `storage.type.fix` for enum values of `FixBind`

Could they instead be called `fix.method`, `fix.conditional`, and `fix.bind`? (Maybe prefixed with something like `keyword.` if it makes sense to group them).

katauber:
> Looks good. Just a bit confused about some names here (do these hook into some VSC logic?):
> 
> * `entity.name.function.fix` for enum values of `FixMethod`
> * `keyword.operator.fix` for enum values of `FixConditional`
> * `storage.type.fix` for enum values of `FixBind`
> 
> Could they instead be called `fix.method`, `fix.conditional`, and `fix.bind`? (Maybe prefixed with something like `keyword.` if it makes sense to group them).

Yes, I know. I am not really happy with this groups either. First I put them in groups which matched the content (see [TextMate Language Grammar](https://macromates.com/manual/en/language_grammars), 12.4 Naming Conventions) BUT then they were styled the same. In the [functional review](https://github.com/metafacture/metafacture-fix/issues/291#issuecomment-1538248910) we decided to color the groups different.
According to the documentation:

> Put the language name last in the name.

Maybe I change the group name suffixes from *.fix to *.method.fix, *.conditional.fix and *.bind.fix? The prefixes would stay as is.

fsteeg:
> Maybe I change the group name suffixes from *.fix to *.method.fix, *.conditional.fix and *.bind.fix? The prefixes would stay as is.

Yes, that sounds good!

PR #299, opened 2023-05-04
Add `entityMemberName` option to Metafix

### LINKED ISSUE #294: Introduce additional option for fix to "repeatedFieldsToEntities" to define `entityName`
I play around with reading all fixPaths of a data stream:

`https://test.metafacture.org/playground/?flux=//+This+part+can+be+adjusted+to+your+own+data+source.+The+example+here+is+marc21%0A%22https%3A//raw.githubusercontent.com/metafacture/metafacture-core/master/metafacture-runner/src/main/dist/examples/read/marc21/10.marc21%22%0A%7C+open-http%0A%7C+as-lines%0A%7C+decode-marc21%0A%0A//+This+part+can+be+reused+for+all+incoming+data+streams+to+list+all+fix+paths.%0A%7C+fix%28%22nothing%28%29%22%2CrepeatedFieldsToEntities%3D%22true%22%29%0A%7C+flatten%0A%7C+encode-json%0A%7C+match%28pattern%3D%22\\.\\d%2B\\.%22%2C+replacement%3D%22.\\%2A.%22%29%0A%7C+match%28pattern%3D%22\\.\\d%2B\%22%22%2C+replacement%3D%22.\\%2A\%22%22%29%0A%7C+decode-json%0A%7C+stream-to-triples%0A%7C+count-triples%28countBy%3D%22predicate%22%29%0A%7C+template%28%22%24%7Bs%7D\t+%24%7Bo%7D%22%29%0A%7C+print%3B%0A%3B%0A&active-editor=fix`

While this works fine for most sources, especially for MARC21 this way is not reliable since indicator subfields can have integers as element names and using 

```
| match(pattern="\\.\\d+\\.", replacement=".\\*.")
| match(pattern="\\.\\d+\"", replacement=".\\*\"")
```

replaces these indicator subfields too and creates therefore wrong fixPaths.

If we could introduce an option to manipulate the result from "repeatedFieldsToEntities" and change the index number to e.g. `"*"`. A similar behaviour is provided by `| decode-json(arrayname="*")` with the option `arrayname`

fsteeg:
e.g. to support outputting array indexes as wildcards when listing available paths. Implemented like `arrayName` in `JsonDecoder`. Added tests in a new class due to NCSS checkstyle warning when expanding `MetafixRecordTest`.

Will resolve #294.


blackwinter (approved):
> Added tests in a new class due to NCSS checkstyle warning when expanding `MetafixRecordTest`.

We've already disabled this check in some cases, maybe it's even unnecessary for test classes.

In general, I'd prefer each class file's tests to live in a correspondingly named test file (`SomeClass.java` -> `SomeClassTest.java`). It's currently not the case and that's sometimes a hindrance to finding a particular test or, even more so, to finding the appropriate place for new tests.

In this particular case: The original tests for `setRepeatedFieldsToEntities()` are still in `MetafixRecordTest.java`, but the new tests for `setEntityMemberName()` (and, coincidentally, for `setRepeatedFieldsToEntities()`) are in `MetafixConfigTest.java`.

I won't object, but I'm feeling a little uneasy about these scattered tests...

fsteeg:
Thanks for the quick review! I've addressed your feedback in https://github.com/metafacture/metafacture-fix/pull/299/commits/935f05fea639137514df3ffe0c7f29b78591681e.

blackwinter:
Great, thanks :)

PR #312, opened 2023-06-13
Update to latest metafacture-core release candidate 

dr0i:
See #311.


dr0i:
If https://github.com/metafacture/metafacture-core/issues/491 is approved we can update the dependency here.

PR #314, opened 2023-06-16
Add uri_encode (#273)

dr0i:
Use metamorph's URLEncoder for uri_encode.

Supersedes https://github.com/metafacture/metafacture-fix/pull/285.

### DIFF
--- a/metafix/src/main/java/org/metafacture/metafix/FixMethod.java
+++ b/metafix/src/main/java/org/metafacture/metafix/FixMethod.java
@@ -21,6 +21,7 @@ import org.metafacture.metafix.maps.RdfMap;
 import org.metafacture.metamorph.api.Maps;
 import org.metafacture.metamorph.functions.ISBN;
 import org.metafacture.metamorph.functions.Timestamp;
+import org.metafacture.metamorph.functions.URLEncode;
 import org.metafacture.metamorph.maps.FileMap;
 
 import java.io.File;
@@ -662,8 +663,16 @@ public enum FixMethod implements FixFunction { // checkstyle-disable-line ClassD
         public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
             record.transform(params.get(0), s -> s.toUpperCase());
         }
+    },
+    uri_encode {
+        @Override
+        public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
+            URL_ENCODER.setPlusForSpace(false);
+            record.transform(params.get(0), URL_ENCODER::process);
+        }
     };
 
+    private static final URLEncode URL_ENCODER = new URLEncode();
     private static final Pattern NAMED_GROUP_PATTERN = Pattern.compile("\\(\\?<(.+?)>");
 
     private static final String FILEMAP_SEPARATOR_OPTION = "sep_char";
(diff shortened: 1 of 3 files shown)


metafix/src/main/java/org/metafacture/metafix/FixMethod.java
```diff
@@ -662,8 +663,16 @@ public void apply(final Metafix metafix, final Record record, final List<String>
         public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
             record.transform(params.get(0), s -> s.toUpperCase());
         }
+    },
+    uri_encode {
+        @Override
+        public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
+            URL_ENCODER.setPlusForSpace(false);
```
blackwinter, line 670: Preferrably, we would expose this as an option `plus_for_space` (along with `safe_chars`). But I don't think it's necessarily a blocker.

metafix/src/main/java/org/metafacture/metafix/FixMethod.java
```diff
@@ -662,8 +663,16 @@ public void apply(final Metafix metafix, final Record record, final List<String>
...
+            record.transform(params.get(0), URL_ENCODER::process);
+        }
     };
 
+    private static final URLEncode URL_ENCODER = new URLEncode();
```
blackwinter, line 675: There is no connection with `NAMED_GROUP_PATTERN`. Please move it to the end, separated by an empty line below `RANDOM`.

_If_ it should be a singleton at all, especially when we're making it configurable (see above).

blackwinter (changes requested):

metafix/src/main/java/org/metafacture/metafix/FixMethod.java
dr0i: Good point. Moved instantiation into method in [c20c2a9](https://github.com/metafacture/metafacture-fix/pull/314/commits/c20c2a945d60fd3a52f3a218bc752b4561069a90).

metafix/src/main/java/org/metafacture/metafix/FixMethod.java
dr0i: Agreed - we can, if needed, easily enhance to use options.

blackwinter:
Would you also document this new function in the README?

metafix/src/main/java/org/metafacture/metafix/FixMethod.java
blackwinter: Okay, but I'll unresolve it as a reminder.

dr0i:
Updated the README.

blackwinter:
Now we're entering shaky territory ;) We differentiate between "record-level" and "field-level" methods. You placed the implementation in the latter section, but the documentation in the former. I don't think we have clear criteria for which is which (@fsteeg, @TobiasNx?), but this one is almost definitely "field-level" I would say (so the documentation should match the implementation).

dr0i:
No, makes sense. Wasn't aware of two sections.  Moved docu snippet  to the proper place. Please have a look again.

blackwinter:
Sorry, but now it's in "script-level" ;) I've moved it down in db84fb7.

dr0i:
Ahhh -  I put it into `Script-level functions` ... will fix that.

blackwinter:
And I named the wrong section in the commit message :( Probably started the weekend a little early ;)

dr0i:
>  Hope that's okay.

timely :)
Going to merge now.

PR #316, opened 2023-06-23
Allow lookup on URLs

### LINKED ISSUE #301: Allow urls with filemap
At the moment one can only refer to a local file via path for `filemap`.

It would be nice if we could support url, so that we could link to file on the web.

`rdfmap` already supports this.

dr0i:
Resolves #301.

Complies with metafacture-core.

- increase max value of ClassDataAbstractionCoupling in checkstyle
- implement test

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -24,7 +24,9 @@
         <module name="CatchParameterName">
             <property name="format" value="^e$"/>
         </module>
-        <module name="ClassDataAbstractionCoupling"/>
+        <module name="ClassDataAbstractionCoupling">
+            <property name="max" value="8"/>
+        </module>
         <module name="ClassFanOutComplexity"/>
         <module name="ClassTypeParameterName"/>
         <module name="ConstantName"/>
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -81,6 +81,9 @@
         <module name="LocalFinalVariableName"/>
         <module name="LocalVariableName"/>
         <module name="MagicNumber"/>
+        <module name="MatchXpath">
+            <property name="query" value="//SINGLE_LINE_COMMENT[./COMMENT_CONTENT[not(starts-with (@text, ' '))]]" />
+        </module>
         <module name="MemberName"/>
         <module name="MethodLength"/>
         <module name="MethodName"/>
--- a/metafix/src/main/java/org/metafacture/metafix/Metafix.java
+++ b/metafix/src/main/java/org/metafacture/metafix/Metafix.java
@@ -36,6 +36,8 @@ import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.UncheckedIOException;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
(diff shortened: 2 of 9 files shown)


config/checkstyle/checkstyle.xml
```diff
@@ -24,7 +24,9 @@
         <module name="CatchParameterName">
             <property name="format" value="^e$"/>
         </module>
-        <module name="ClassDataAbstractionCoupling"/>
+        <module name="ClassDataAbstractionCoupling">
+            <property name="max" value="8"/>
```
blackwinter, line 28: That's one way to do it, sure. But then you should also review where this check has been disabled previously and whether these suppressions can be removed.

Alternatively, disable the check where it's failing now. We're going to review the whole Checkstyle configuration at some point.

metafix/src/main/java/org/metafacture/metafix/Metafix.java
```diff
@@ -148,26 +150,43 @@ public void literal(final String name, final String value) {
     }
 
     public String resolvePath(final String path) {
-        final Path basePath;
+        String resolvedPath = path;
 
-        if (path.startsWith(".")) {
-            if (fixFile != null) {
-                basePath = getPath(fixFile).getParent();
+        if (isValidUrl(path)) {
```
blackwinter, line 155: Why not assign `resolvedPath` here?

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
```diff
@@ -67,20 +70,27 @@ public MetafixLookupTest() {
...
-
+        String responseBody = loadFile(RDF_PATH);
         WIRE_MOCK_SERVER.stubFor(WireMock.get(urlPatternRedirectToUrl)
             .willReturn(WireMock.aResponse()
                 .withHeader("Content-Type", "text/turtle")
                 .withBody(responseBody)));
+
+        //stub for CSV
+        urlPattern = WireMock.urlPathEqualTo(CSV_PATH);
+        responseBody = loadFile(CSV_PATH);
```
blackwinter, line 87: Why not give these variables appropriate names?

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
```diff
@@ -67,20 +70,27 @@ public MetafixLookupTest() {
...
-        final String responseBody = new BufferedReader(new InputStreamReader(
-                    MetafixLookupTest.class.getResourceAsStream("." + RDF_PATH))).lines().collect(Collectors.joining("\n"));
-
+        String responseBody = loadFile(RDF_PATH);
         WIRE_MOCK_SERVER.stubFor(WireMock.get(urlPatternRedirectToUrl)
             .willReturn(WireMock.aResponse()
                 .withHeader("Content-Type", "text/turtle")
                 .withBody(responseBody)));
+
+        //stub for CSV
```
blackwinter, line 85: Minor: Missing space after comment marker.

blackwinter (changes requested):
First impression, not the final review...

In general, I'd _really_ prefer variables to be kept `final` as much as possible. Could you please try to adapt your changes?

config/checkstyle/checkstyle.xml
dr0i: Good advise - wasn't aware of that possibility. Disabled the check.

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
dr0i: OK, I've tried to do this by not reusing Objects. I hope that was your intention here.

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
dr0i: Added a checkstyle rule an btw updated this dependency. This brought up a some [code commented out](https://github.com/metafacture/metafacture-fix/pull/316/commits/8500ac4392a8760e313013a431bfdfc63aa9ed99) which I did in part just removed and in part left it (just "fixing" the structure the new checkstyle rule demands) as these comments may have some worth. I didn't find any remarks in git commits so I leave this to you to decide. Actually I don't know if you appreciate the new checkstyle rule at all or want to dump that.

blackwinter: > Actually I don't know if you appreciate the new checkstyle rule at all or want to dump that.

In fact, I don't, sorry. Exactly because it can't differentiate between comments and commented-out code. Some things just have to be left to human judgement ;)

metafix/src/main/java/org/metafacture/metafix/Metafix.java
blackwinter: > That demands the use of `path` to not have an initialized variable.

Can you elaborate? There is no change w.r.t. the `path` variable.

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
blackwinter: Yes, thanks, that's what I had in mind. Although I don't understand this part:

> also not reusing Objects.

dr0i:
I opened new #317 to consolidate commits into one.

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
dr0i: In https://github.com/metafacture/metafacture-fix/pull/316/commits/eef131aaba7cdfaae44a34e6deccaafc80386b56 both objects `urlPattern` and `responseBody` where reused. That made the naming rather generic.

metafix/src/main/java/org/metafacture/metafix/Metafix.java
dr0i: I cannot assign `final String resolvedPath = path` to use `resolvedPath` in `if (isValidUrl(path)) {`, right? Maybe I just don't understand what you meant here.

metafix/src/test/java/org/metafacture/metafix/MetafixLookupTest.java
dr0i: OK. Dumped that in #317.

blackwinter:
Why am I assigned here? Is this PR superseded by #317 or not?

dr0i:
I assigned you because of some new comments from me here I though you may close by clicking on `resolve conversation`. But as you were fine with the superseding PR I close this.

PR #337, opened 2023-11-21
Fix function "uri_encode" to be backwards compatible

dr0i:
See  metafacture/metafacture-fix#273.

- introduce options "safe_chars" and "plus_for_space" for function "uri_encode"
- update README
- bump metafacture dependency to 5.7.0-rc2

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,9 +717,10 @@ Encodes a field value as URI. Aka percent-encoding.
 
 Options:
 
-- `allow_empty_values`: Sets whether to allow empty values in the filemap or to ignore these entries. (Default: `false`)
-- `compression`: Sets the compression of the file.
-
+- `plus_for_space`: Sets whether "space" (' ') will be substituted by a "plus" ('+') or be percent escaped ('%20'). Default: `true`)
+- `safe_chars`: Sets characters that won't be escaped. Safe characters are the
+ranges 0..9, a..z and A..Z. These are always safe and should not be specified.
+Default safe characters are also ".", "-", "*", and "_".
 
 ```perl
 uri_encode("<sourceField>"[, <options>...])
--- a/README.md
+++ b/README.md
@@ -717,7 +717,7 @@ Encodes a field value as URI. Aka percent-encoding.
 
 Options:
 
-- `plus_for_space`: Sets whether "space" (' ') will be substituted by a "plus" ('+') or be percent escaped ('%20'). Default: `true`)
+- `plus_for_space`: Sets whether "space" (` `) will be substituted by a "plus" (`+`) or be percent escaped (`%20`). (Default: `true`)
 - `safe_chars`: Sets characters that won't be escaped. Safe characters are the
 ranges 0..9, a..z and A..Z. These are always safe and should not be specified.
 Default safe characters are also ".", "-", "*", and "_".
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -715,8 +715,20 @@ upcase("<sourceField>")
 
 Encodes a field value as URI. Aka percent-encoding.
 
+Options:
+
+- `allow_empty_values`: Sets whether to allow empty values in the filemap or to ignore these entries. (Default: `false`)
+- `compression`: Sets the compression of the file.
```
blackwinter, line 721: These are not options for `uri_encode()`.

metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
```diff
@@ -4015,12 +4015,48 @@ public void shouldUriEncodePathSegment() {
...
+                o.get().literal("id", "%2FDE-A96%3A%25%20%283%29%23%21");
+                o.get().endRecord();
+            }
+        );
+    }
+
+    @Test
+    public void shouldUriEncodePathSegmentWithoutSafeChars() {
+        MetafixTestHelpers.assertFix(streamReceiver, Arrays.asList(
+                "uri_encode('id', safe_chars:'', plus_for_space:'false')"
```
blackwinter, line 4050: Drop `plus_for_space` option as this is supposed to test `safe_chars`.

blackwinter (changes requested):

README.md
dr0i: cheeze - thx!

metafix/src/test/java/org/metafacture/metafix/MetafixMethodTest.java
dr0i: Did this now, see 9bf530703ceeb8d80283a6855bc3d4c43e6efaab.

README.md
```diff
@@ -715,8 +715,21 @@ upcase("<sourceField>")
 
 Encodes a field value as URI. Aka percent-encoding.
 
+Options:
+
+- `plus_for_space`: Sets whether "space" (' ') will be substituted by a "plus" ('+') or be percent escaped ('%20'). Default: `true`)
```
blackwinter, line 720: Use Markdown formatting for literal values and add missing parentheses before "Default":

```suggestion
- `plus_for_space`: Sets whether "space" (` `) will be substituted by a "plus" (`+`) or be percent escaped (`%20`). (Default: `true`)
```

README.md
```diff
@@ -715,8 +715,21 @@ upcase("<sourceField>")
...
+- `safe_chars`: Sets characters that won't be escaped. Safe characters are the
+ranges 0..9, a..z and A..Z. These are always safe and should not be specified.
+Default safe characters are also ".", "-", "*", and "_".
```
blackwinter, line 723: Should be on one line and maintain format for default value:

```suggestion
- `safe_chars`: Sets characters that won't be escaped. Safe characters are the ranges 0..9, a..z and A..Z. These are always safe and should not be specified. (Default: `.-*_`)
```

blackwinter (changes requested):

dr0i:
Committed your  suggestions  - thx!

blackwinter (approved):

PR #353, opened 2024-05-24
Reuse the Metafacture Core .gitignore

TobiasNx:
https://github.com/metafacture/metafacture-core/blob/master/.gitignore

It is nicely formatted and added the files from the old .gitignore that were missing.


blackwinter (changes requested):
This would ignore `misc/bin/` which is a legitimate directory.

In general, I would prefer local peculiarities to be handled locally, i.e. `.git/info/exclude` (or global ignore file via `core.excludesfile` config).

In addition to that, the list seems include a lot of outdated stuff.

fsteeg:
> In general, I would prefer local peculiarities to be handled locally, i.e. `.git/info/exclude`

Yes, makes sense. @TobiasNx here are some details on that:
https://github.com/github/docs/blob/main/content/get-started/getting-started-with-git/ignoring-files.md#excluding-local-files-without-creating-a-gitignore-file

dr0i:
Wasn't aware of ` .git/info/exclude` - good thing :+1:

TobiasNx:
then closing

PR #356, opened 2024-06-06
Use the default option as in catmandu #333

TobiasNx:
Resolves #333.

Issue #354, opened 2024-05-28
Add function to transform string to base64 hash


TobiasNx:
In context of OERSI we need to transform a string to base64.

TobiasNx:
Yes

PR #365, opened 2024-09-18
Add new fix method to set variable with element value #282

### LINKED ISSUE #282: Create variables from element values or names.
At the moment you can only create variables from explicitly assigning values to variable names by `put_var` or `put_vars`.

I suggest a `to_var` function:

 `to_var("sourceElement","variableName",variableValue:"key/value")`

Perhaps the feature for element names is not needed.

TobiasNx:
Resolves #282

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,16 +269,15 @@ put_vars(
 
 ##### `to_var`
 
-Defines a single global variable that can be referenced with `$[<variableName>]` and assinges the value of the `<sourceField>`.
+Defines a single global variable that can be referenced with `$[<variableName>]` and assigns the value of the `<sourceField>`.
 
 ```perl
-to_var("<sourceField>", "<variableValue>")
+to_var("<sourceField>", "<variableName>")
 ```
 
 Options:
 
-- `"default"`: provide default value if source field does not exist.
-
+- `default`: Default value if source field does not exist. The option needs to be written in quotation marks because it is a reserved word in Java. (Default: Empty string)
 
 [Example in Playground](https://metafacture.org/playground/?example=to_var)
 
--- a/README.md
+++ b/README.md
@@ -267,6 +267,23 @@ put_vars(
 
 [Java Code](https://github.com/search?type=code&q=repo:metafacture/metafacture-fix+path:FixMethod.java+"+put_vars+{")
 
+##### `to_var`
+
+Defines a single global variable that can be referenced with `$[<variableName>]` and assinges the value of the `<sourceField>`.
+
+```perl
+to_var("<sourceField>", "<variableValue>")
+```
+
+Options:
+
+- `"default"`: provide default value if source field does not exist.
+
+
+[Example in Playground](https://metafacture.org/playground/?example=to_var)
+
+[Java Code](https://github.com/search?type=code&q=repo:metafacture/metafacture-fix+path:FixMethod.java+"+to_var+{")
+
 #### Record-level functions
 
 ##### `add_field`
--- a/metafix/src/main/java/org/metafacture/metafix/FixMethod.java
+++ b/metafix/src/main/java/org/metafacture/metafix/FixMethod.java
@@ -127,7 +127,7 @@ public enum FixMethod implements FixFunction { // checkstyle-disable-line ClassD
         @Override
         public void apply(final Metafix metafix, final Record record, final List<String> params, final Map<String, String> options) {
             final Value value = record.get(params.get(0));
-            metafix.getVars().put(params.get(1), Value.isNull(value) ? options.getOrDefault(DEFAULT_OPTION, "") : value.toString());
+            metafix.getVars().put(params.get(1), Value.isNull(value) ? options.getOrDefault(DEFAULT_OPTION, "") : value.asString());
         }
     },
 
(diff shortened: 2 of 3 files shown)


blackwinter (changes requested):
I've updated the implementation to account for missing values and added tests for edge cases (22e0482).

Open questions:

1. How to handle missing values? Is an empty string the right choice?
2. How to handle non-string values? Is the Java object representation the right choice? (Previously, you [said](https://github.com/metafacture/metafacture-fix/issues/282#issuecomment-2355583039) that it "should only work with string-elements".)

Finally, the new function needs to be documented.

TobiasNx:
> I've updated the implementation to account for missing values and added tests for edge cases ([a52dc9b](https://github.com/metafacture/metafacture-fix/commit/a52dc9bd3c897d7c6461ba6846e3b2819faa653b)).
> 
> Open questions:
> 
>     1. How to handle missing values? Is an empty string the right choice?
I would not add a variable if the element does not exist.

> 
>     2. How to handle non-string values? Is the Java object representation the right choice? (Previously, you [said](https://github.com/metafacture/metafacture-fix/issues/282#issuecomment-2355583039) that it "should only work with string-elements".)
I thought MF only processes string values in the Stream-Processing. Because of that we introduced marker for numbers in the json-encoder. The idea with string was the connection to the other ticket in mf-core but I have put the idea in core on hold.

> 
> 
> Finally, the new function needs to be documented.

blackwinter:
> I thought MF only processes string values in the Stream-Processing.

Metafix knows data types strings, arrays and objects; variables can only contain strings.

TobiasNx:
> > I thought MF only processes string values in the Stream-Processing.
> 
> Metafix knows data types strings, arrays and objects; variables can only contain strings.

Ah, I forgot. Then I would suggest an exception that the function expected a string but got hash or array.

blackwinter:
> I would not add a variable if the element does not exist.

This might become difficult to use:

1. If one record has the field and the next one doesn't, the variable keeps its previous value. We might want to remove the variable instead.
2. Referencing a non-existing variable (if we remove it or if it just hasn't been set yet) results in an error. So you can't reliably use the variable. We might have to introduce a conditional that explicitly checks for the presence of a variable.

TobiasNx:
> > I would not add a variable if the element does not exist.
> 
> This might become difficult to use:
> 
>     1. If one record has the field and the next one doesn't, the variable keeps its previous value. We might want to remove the variable instead.
> 
>     2. Referencing a non-existing variable (if we remove it or if it just hasn't been set yet) results in an error. So you can't reliably use the variable. We might have to introduce a conditional that explicitly checks for the presence of a variable; or add an option to specify the default value.

I favor the the option to define a default value. The conditional for the a variable sounds good. I am also in favor to delete the variable at the end of each record.

blackwinter:
> Then I would suggest an exception that the function expected a string but got hash or array.

Implemented in cee638a.

> I favor the the option to define a default value.

Implemented in 46d1ef5.

> The conditional for the a variable sounds good. I am also in favor to delete the variable at the end of each record.

Should probably be handled in separate issues.

TobiasNx:
@TobiasNx I added documentation. Should @fsteeg do a code review?

README.md
```diff
@@ -267,6 +267,23 @@ put_vars(
...
+
+Defines a single global variable that can be referenced with `$[<variableName>]` and assinges the value of the `<sourceField>`.
+
+```perl
+to_var("<sourceField>", "<variableValue>")
+```
+
+Options:
+
+- `"default"`: provide default value if source field does not exist.
```
blackwinter, line 280: Drop quotes, capitalize first word after colon.

Option description should say what it _is_, not what it _does_.

```suggestion
- `default`: Default value if source field does not exist.
```

README.md
```diff
@@ -267,6 +267,23 @@ put_vars(
...
+
+
```
blackwinter, line 282: Extra empty line.

blackwinter (changes requested):

README.md
```diff
@@ -267,6 +267,23 @@ put_vars(
 
 [Java Code](https://github.com/search?type=code&q=repo:metafacture/metafacture-fix+path:FixMethod.java+"+put_vars+{")
 
+##### `to_var`
+
+Defines a single global variable that can be referenced with `$[<variableName>]` and assinges the value of the `<sourceField>`.
```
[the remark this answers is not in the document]
TobiasNx:
```suggestion
Defines a single global variable that can be referenced with `$[<variableName>]` and assigns the value of the `<sourceField>`.
```

README.md
```diff
@@ -267,6 +267,23 @@ put_vars(
...
+
+```perl
+to_var("<sourceField>", "<variableValue>")
```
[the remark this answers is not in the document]
TobiasNx:
```suggestion
to_var("<sourceField>", "<variableName>")
```

README.md
TobiasNx: Isn't it important to highlight that this option has to be in quotes?

blackwinter: See [discussion in #354](https://github.com/metafacture/metafacture-fix/pull/356#discussion_r1799506405).

README.md
TobiasNx:
```suggestion
```

README.md
TobiasNx:
```suggestion
- `default`: Default value if source field does not exist. The option needs to be written in quotation marks because it is a reserved word in Java. (Default: Empty string)
```

README.md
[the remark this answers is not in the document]
TobiasNx: Done. Thanks.

README.md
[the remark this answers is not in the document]
TobiasNx: Done. Thanks.

blackwinter (approved):

fsteeg (approved):
Very useful feature!

Issue #309, opened 2023-06-06
Change `set_field()` behaviour to not create intermediate structures.


blackwinter:
`add_field()` and `set_field()` are going to equivalent with #308, but according to [Catmandu's documentation](https://metacpan.org/pod/Catmandu::Fix::set_field) `set_field()` "will not create the intermediate structures if they are missing".

Functional review: @TobiasNx
Code review: @blackwinter / @fsteeg

TobiasNx:
The aspect of the intermediate structure seems to be true for all fixes of the `set_`-family so not just for `set_field` but also for `set_array` and `set_hash`. I think that the adjustment for all these fixes makes sense since we want to adjust to the behaviour of Catmandu BUT on the side of lobid and oersi we rely on a fixes that create empty hashes and empty arrays WITH the intermediate structure. So If we move on with this, we should also introduce two new fix functions `add_array` and `add_hash` as counterpart to `set_hash` and `set_array`.

TobiasNx:
`add_array` and `add_hash` are introduced now and the set functions are only aliases now.
We should advise anybody to use the add functions instead of the set ones until the new behaviour is introduced.

Also wenn the behaviour change is implemented we should rename all set_ tests to add_ ones as well as add new set functions.

PR #379, opened 2024-10-16
Add new functions add_array and add_hash #374

### LINKED ISSUE #374: Introduce `add_hash` and `add_array` fixes
This is a prerequisite of #309 .
In Catmandu all `set_`-functions do not create intermediate structure.

Since we are relying heavily in context of OERSI and lobid on creating the the intemediate structure with `set_array`and `set_hash` . To add `add_array` and `add_hash` as copy of the current implementation of `set_hash` and `set_array` would be an easy helper so that we would not have to change the transformation drastically.

TobiasNx:
Resolves #374

This is a preparation for #309 which will change the behavior of all `set-`functions as `set_array` and `set_hash` which will not create the intermediate structure of a path.

Since we rely in context of oersi and lobid on this specific functionality I introduced `add_array` and `add_hash` which copy the current implementation of `set_array` and `set_hash`.

I also changed the tests from set_array and set_hash to add_array and add_hash since #307 will need new integration tests and java tests.


blackwinter (changes requested):
I would suggest to change all `set_*` implementations to delegate to the corresponding `add_*` functions, so we don't have to maintain two identical implementations for each pair until the former have been replaced.

I would also argue that we should keep `set_*` where appropriate, because these are the "standard" functions; the new functions aren't supported by Catmandu (yet).

Similarly, all the tests should be left untouched in order to surface any actual behaviour changes once the `set_*` implementations are changed. Just add new ones where appropriate.

fsteeg (commented):
Agree about suggestions by @blackwinter.

TobiasNx:
@blackwinter like this?

blackwinter (approved):

Issue #383, opened 2024-11-07
Enable the runner to select debug level.


TobiasNx:
> Also `metafacture-runner/src/main/dist/config/java-options.conf`.
> It seems that the core included runner can adjust this. 
> BUT the metafix runner cannot do this.

_Originally posted by @TobiasNx in https://github.com/metafacture/metafacture-core/issues/536#issuecomment-2462095308_
            

The Metafix Runner is not able to set the config level to DEBUG.

blackwinter:
See my earlier [comment in #371](https://github.com/metafacture/metafacture-fix/issues/371#issuecomment-2419760893).

TobiasNx:
So we have to sit this one out until core and fix are merged?

TobiasNx:
Creating a config folder in the bin folder with the log4j.xml does not do the trick. @fsteeg suggested it to try it our.

TobiasNx:
Does it make sense to adjust the runner build to the core one? There are a lot of advantages to this. Since the runner would have examples and more.

dr0i:
You may want to cherry-pick [47304f9](https://github.com/metafacture/metafacture-fix/commit/47304f96af35f65dfc95b6c45fa04bed0455b1b2) and log something at debug level.

TobiasNx:
The following seems to work:

```
$ ./gradlew :metafix-runner:run --args="/home/tobias/git/metafacture-core/metafacture-runner/build/install/metafacture-core/examples/morph/marc21/morph-marc21.flux"

> Task :metafix-runner:run
17:18:09.245 [main] DEBUG org.metafacture.monitoring.StreamLogger -- start record 946638705
17:18:09.261 [main] DEBUG org.metafacture.monitoring.StreamLogger -- literal dc:title=Colonial and post-colonial discourse in the novels of Yo§am Sang-So§ap, Chinua Achebe and Salman Rushdie
Colonial and post-colonial discourse in the novels of Yo§am Sang-So§ap, Chinua Achebe and Salman Rushdie17:18:09.265 [main] DEBUG org.metafacture.monitoring.StreamLogger -- end record
17:18:09.265 [main] DEBUG org.metafacture.monitoring.StreamLogger -- start record 94685887X
17:18:09.267 [main] DEBUG org.metafacture.monitoring.StreamLogger -- literal dc:title=Ostenfelder Bauernhaus : Deutschlands a§�ltestes Freilichtmuseum in Husum

[...]

PR #387, opened 2024-11-22
Add log4j logging and bridge

dr0i:
Enables logging.

See  #383.

### DIFF
--- a/metafix-runner/build.gradle
+++ b/metafix-runner/build.gradle
@@ -30,6 +30,9 @@ dependencies {
   implementation "org.metafacture:metafacture-xml:${versions.metafacture}"
   implementation "org.metafacture:metafacture-yaml:${versions.metafacture}"
   implementation "org.metafacture:metamorph:${versions.metafacture}"
+  runtimeOnly 'org.apache.logging.log4j:log4j-core'
+  implementation 'org.apache.logging.log4j:log4j-1.2-api:2.9.1'
+  implementation 'ch.qos.logback:logback-classic:1.3.14'
 
   implementation('org.antlr:antlr-runtime') {
     version { strictly versions.antlr }
--- a/metafix/src/main/resources/log4j.xml
+++ b/metafix/src/main/resources/log4j.xml
@@ -1,0 +1,18 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE log4j:configuration SYSTEM "http://logging.apache.org/log4j/1.2/apidocs/org/apache/log4j/xml/doc-files/log4j.dtd">
+
+<log4j:configuration>
+    <appender name="stdout" class="org.apache.log4j.ConsoleAppender">
+        <layout class="org.apache.log4j.PatternLayout">
+            <param name="ConversionPattern"
+                value="%-5p [%t] [%c{1}] %m%n" />
+        </layout>
+    </appender>
+
+    <root>
+        <priority value="INFO" />
+        <appender-ref ref="stdout" />
+    </root>
+
+</log4j:configuration>
+
--- a/metafix/src/main/resources/log4j.xml
+++ b/metafix/src/main/resources/log4j.xml
@@ -10,7 +10,7 @@
     </appender>
 
     <root>
-        <priority value="INFO" />
+        <priority value="${org.metafacture.metafix.logLevel}" />
         <appender-ref ref="stdout" />
     </root>
 


metafix-runner/build.gradle
```diff
@@ -30,6 +30,9 @@ dependencies {
   implementation "org.metafacture:metafacture-xml:${versions.metafacture}"
   implementation "org.metafacture:metafacture-yaml:${versions.metafacture}"
   implementation "org.metafacture:metamorph:${versions.metafacture}"
+  runtimeOnly 'org.apache.logging.log4j:log4j-core'
+  implementation 'org.apache.logging.log4j:log4j-1.2-api:2.9.1'
+  implementation 'ch.qos.logback:logback-classic:1.3.14'
```
blackwinter, line 35:
```suggestion
  runtimeOnly "org.slf4j:slf4j-log4j12:${versions.slf4j}"
```

blackwinter (changes requested):
Why are you introducing `log4j`, and also `logback`? We're already using `slf4j`.

blackwinter:
> Tested with dist created with `installDist`, ran flux with `log-object`, got output 👍

But not without increasing the log level first, right?

metafix-runner/build.gradle
fsteeg: With this suggestion, I get no logging output, and the same `Class path contains multiple SLF4J bindings.` warning that I get without this PR.

fsteeg:
> But not without increasing the log level first, right?

Oh, right, weird. I saw the original commit https://github.com/metafacture/metafacture-fix/commit/47304f96af35f65dfc95b6c45fa04bed0455b1b2 and thought it makes sense, but the current version is actually INFO, but I still get the DEBUG output, without any changes to `log4j.xml`.

metafix-runner/build.gradle
blackwinter: That warning aside (different issue, IMO), you would have to increase the log level to `DEBUG`.

blackwinter:
> but I still get the DEBUG output, without any changes to `log4j.xml`.

I see. Now that you mention it, the log format is also different than what is specified in the config. Could it be that logback ignores the config file?

blackwinter:
That's another reason why I think we shouldn't introduce any new logging dependencies. metafacture-runner uses `slf4j-log4j` and so should metafix-runner.

metafix/src/main/resources/log4j.xml
```diff
@@ -0,0 +1,18 @@
...
+<log4j:configuration>
+    <appender name="stdout" class="org.apache.log4j.ConsoleAppender">
+        <layout class="org.apache.log4j.PatternLayout">
+            <param name="ConversionPattern"
+                value="%-5p [%t] [%c{1}] %m%n" />
+        </layout>
+    </appender>
+
+    <root>
+        <priority value="INFO" />
```
blackwinter, line 13: Simplify changing the log level at runtime:

```suggestion
        <priority value="${env:METAFIX_LOG_LEVEL:-INFO}" />
```

metafix-runner/build.gradle
fsteeg: Same with `DEBUG` in `log4j.xml`: no logging output. As you wrote below, it must be use some other config?

blackwinter: That's not what I saw. Changing the log level (priority) in `log4j.xml` and then running `./gradlew installDist` yields debug output with `log-object`.

fsteeg: Weird, I have `<priority value="DEBUG" />` in `metafix/src/main/resources/log4j.xml`, only `runtimeOnly "org.slf4j:slf4j-log4j12:${versions.slf4j}"` added in `metafix-runner/build.gradle`: no logging output for `log-object`.

blackwinter: And you ran `./gradlew installDist` afterwards? For me on `master`:

1. Copy `log4j.xml` from metafacture-core to `metafix/src/main/resources/`.
2. Change priority `info` to `DEBUG` in `metafix/src/main/resources/log4j.xml`.
3. Add `runtimeOnly "org.slf4j:slf4j-log4j12:${versions.slf4j}"` to `metafix-runner/build.gradle`.
4. Run `./gradlew installDist`.
5. Change into `metafix-runner/build/install/metafix-runner` directory.
6. Run `bin/metafix-runner /path/to/flux-with-log-object`.
7. See "DEBUG [main] ..." output.

dr0i: @blackwinter your proposal is working for me

metafix/src/main/resources/log4j.xml
dr0i: How is the variable set? `export METAFIX_LOG_LEVEL=DEBUG; bin/metafix-runner  $pathTo.flux` didnt' work. Same for `export METAFIX_LOG_LEVEL="DEBUG"`.

dr0i: hmpf, no this works , BUT it's always on DEBUG level regardless what I set .... oh , logging is so complicated ...

blackwinter: Oh, sorry, I overlooked that we're on Log4j 1.x which doesn't seem to support [property substitution](https://logging.apache.org/log4j/2.x/manual/configuration.html#property-substitution).

blackwinter: No, it doesn't work. It's effectively an invalid/empty log level.

metafix-runner/build.gradle
fsteeg: I just tried it on my home machine, and here it works. Maybe some cached dependency bringing in some other config file on my office machine? So that would be +1 for using only `runtimeOnly "org.slf4j:slf4j-log4j12:${versions.slf4j}"`. However, it works here even without adding any dependency, just by switching `INFO` to `DEBUG` in `metafix/src/main/resources/log4j.xml`.

blackwinter: > However, it works here even without adding any dependency

Indeed. Probably because it's pulled in by metafacture-runner.So let's just drop it altogether then.

metafix/src/main/resources/log4j.xml
blackwinter: Okay, apparently it does support property substitution, it's just [not readily documented](https://logging.apache.org/log4j/1.x/apidocs/org/apache/log4j/PropertyConfigurator.html). However, I wasn't able to figure out how to supply a default value in the configuration file. So it gets a little more complicated:

```xml
<priority value="${org.metafacture.metafix.logLevel}" />
```

```diff
diff --git metafix-runner/build.gradle metafix-runner/build.gradle
index 1179b5c..ab71abb 100644
--- metafix-runner/build.gradle
+++ metafix-runner/build.gradle
@@ -39,11 +39,15 @@ dependencies {
 application {
   mainClass = 'org.metafacture.runner.Flux'
 
+  applicationDefaultJvmArgs = [
+    "-Dorg.metafacture.metafix.logLevel=INFO"
+  ]
+
   if (project.hasProperty('profile')) {
     def file = project.getProperty('profile') ?: project.name
     def depth = project.hasProperty('profile.depth') ? project.getProperty('profile.depth') : 8
 
-    applicationDefaultJvmArgs = [
+    applicationDefaultJvmArgs += [
       "-XX:FlightRecorderOptions=stackdepth=${depth}",
       "-XX:StartFlightRecording=dumponexit=true,filename=${file}.jfr,settings=profile"
     ]
```

Then run with `JAVA_OPTS=-Dorg.metafacture.metafix.logLevel=DEBUG`.

dr0i:
Closing this PR in favour of https://github.com/metafacture/metafacture-fix/pull/388 . Thx for all the valuable input here!

PR #388, opened 2024-11-28
Add log4j logging

dr0i:
Enables java logging. The log level can be changed by setting a variable. Set and run like this:
JAVA_OPTS="-Dorg.metafacture.metafix.logLevel=DEBUG"; ./gradlew installDist;  cd metafix-runner/build/install/metafix-runner; bin/metafix-runner $pathTo.flux

See #383.

Substitutes #387.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,12 @@ To execute a Fix (embedded in a Flux) via CLI:
 
 `./gradlew :metafix-runner:run --args="$PWD/path/to.flux"`
 
+To execute a Fix (embedded in a Flux) via CLI in java debug mode:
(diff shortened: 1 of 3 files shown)


blackwinter (approved):
The usage instructions are wrong, though. You either have to _export_ the `JAVA_OPTS` variable (which means it applies during the whole shell session, until reset or unset) or _prefix_ the runner command with it:

1. `export JAVA_OPTS=...; bin/metafix-runner ...`
2. `JAVA_OPTS=... bin/metafix-runner ...`

Also note that one can't change the log level this way when running directly via Gradle (`./gradlew :metafix-runner:run ...`), it only works _within_ the distribution.

PR #391, opened 2025-03-07
Fix memory leak (metafacture-core#666)

dr0i:
DON'T MERGE THIS PR!

Fixes a memory leak for old  programs running java 8.

This is a branch which fixes a memory leak (see  (metafacture-core#666)). It's build on the latest release which uses java 8.

Resolves https://github.com/metafacture/metafacture-core/issues/666#issuecomment-2706080485.

Use like:
look at the needed metafacture-core 5.7.1 dependency build as noted in https://github.com/metafacture/metafacture-fix/blob/publish-0.7.1-fixMemoryLeak/.github/workflows/build.yml. Then:
`git checkout 0.7.1;  /gradlew publishToMavenLocal`.  You can then consume the locally build `metafix/0.7.1`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,6 +13,14 @@ jobs:
         java-version: 1.8
     - name: Grant execute permission for gradlew
       run: chmod +x gradlew
+    - name: Install metafacture-core 5.7.1
+      run: |
+        cd ..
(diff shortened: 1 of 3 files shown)


dr0i:
Closed intentionally. This branch is needed because we cannot make a release because we cannot release the needed metafacture-core 5.7.1 version atm.

